Issue #11, opened 2022-11-20
Run the tests in GitHub Actions


simonw:
It looks like CI isn't up and running yet. Would you be interested in a PR that sets that up with GitHub Actions?

I run CI against a Django app that uses PostgreSQL in https://github.com/natbat/pillarpointstewards/blob/main/.github/workflows/test.yml - I can adapt that for this repo if it would be useful.

andrewgodwin:
I was waiting to set that up myself once I have more tests, but if you want to get me a ready-made one I'll take it!

simonw:
PR:
- #13

PR #13, opened 2022-11-20
Configure GitHub Actions to run tests

simonw:
This starts a PostgreSQL service (using [this pattern](https://til.simonwillison.net/github-actions/postgresq-service-container)) and runs the tests against it.

I had to mark one of the tests as `xfail` because it's exercising a URL which I don't think has been implemented yet (`/@test@example.com/actor/`).

Refs:
- #11

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,35 @@
+name: Test
+
+on: [push, pull_request]
+
+jobs:
+  test:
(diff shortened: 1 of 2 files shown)


andrewgodwin (approved):

andrewgodwin:
Hilariously this is because I moved the actor URLs to be just `@username` yesterday and forgot to run the tests, so that's actually a valid fail!

PR #16, opened 2022-11-20
Add missing Debian build dependencies for Python 3 dependencies

joshdick:
Python dependencies failed to build/install as part of the Docker image build until I added these Debian dependencies to the Dockerfile.

Was able to reproduce the issue before this change on a 2021 M1 MacBook Pro running macOS Ventura 13.0.1 and Docker Desktop 4.14.1 (91661) / Engine 20.10.21.


andrewgodwin:
Ah, fascinating, this must be an architecture difference where there's x64 wheels for some packages and not ARM ones. I do want to run takahe.social on ARM processors soonish so this is handy!

Issue #17, opened 2022-11-20
Docker use is... strange


danielquinn:
On most projects, the `Dockefile` lives in the project root alongside `docker-compose.yaml` which, while annoying as it clutters up the root, makes actually using it quite easy:

```shell
$ git clone ...
cd takahe
docker-compose up
```

However in its current position living in `docker/`, it's not clear what I'm supposed to do there.  If I hop into that folder and run `docker-compose up`, it barfs, complaining that `requirements.txt` isn't available:

```shell
$ docker-compose up
[+] Running 9/10
 ⠿ web Warning                                                                                                                                                                1.7s
 ⠿ db Pulled                                                                                                                                                                 23.7s
   ⠿ ca7dd9ec2225 Pull complete                                                                                                                                               1.1s
   ⠿ 126ca7716c09 Pull complete                                                                                                                                               1.7s
   ⠿ 72a45366ffb3 Pull complete                                                                                                                                               1.9s
   ⠿ c4825ba03629 Pull complete                                                                                                                                              20.1s
   ⠿ 90d9e7b7d99d Pull complete                                                                                                                                              20.6s
   ⠿ a6f996156d58 Pull complete                                                                                                                                              20.8s
   ⠿ 4af28305ac48 Pull complete                                                                                                                                              21.0s
   ⠿ aae4e1126571 Pull complete                                                                                                                                              21.1s
[+] Building 1.0s (7/11)                                                                                                                                                           
 => [internal] load build definition from Dockerfile                                                                                                                          0.1s
 => => transferring dockerfile: 563B                                                                                                                                          0.0s
 => [internal] load .dockerignore                                                                                                                                             0.1s
 => => transferring context: 2B                                                                                                                                               0.0s
 => [internal] load metadata for docker.io/library/python:3.11.0-slim-buster                                                                                                  0.6s
 => [1/7] FROM docker.io/library/python:3.11.0-slim-buster@sha256:e3168357ada3287cda4c79ecae7f5103cb4d071d9b610a1d9380ea4b7cfc981f                                            0.2s
 => => resolve docker.io/library/python:3.11.0-slim-buster@sha256:e3168357ada3287cda4c79ecae7f5103cb4d071d9b610a1d9380ea4b7cfc981f                                            0.1s
 => [internal] load build context                                                                                                                                             0.1s
 => => transferring context: 1.11kB                                                                                                                                           0.0s
[...]

PR #21, opened 2022-11-20
Document Docker development

danielquinn:
While I'd argue that moving the Docker files into the project root is the more accessible option, this PR at least documents how to make use of them where they are.  See #17 for reference.

This also mounts the project root at /takahe/ to make development possible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@
 /docs/_build
 /media/
 notes.md
+__pycache__/
+*.pyc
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -5,3 +5,5 @@
 /docs/_build
 /media/
 notes.md
+__pycache__/
+*.pyc
```
danielquinn, line 9: Perhaps leveraging one of GitHub's .gitignore templates would be useful here.

andrewgodwin: I have personal templates I just didn't copy over yet due to, well, this all being very alpha :) 

(my systems all have global ignores configured)

PR #25, opened 2022-11-20
(fix) mimetype for JPEG file extensions

lullis:
I started reading through the code and found the lowest hanging fruit possible. :)


andrewgodwin:
Nice catch!

PR #26, opened 2022-11-20
Enforce signups_allowed=False

manfre:
Directly accessing the signup page when `signups_allowed=False` will no longer show the form fields and "create" button. Added form validation check as extra safety against direct POST.


users/tests/views/test_auth.py
```diff
@@ -0,0 +1,59 @@
...
+from core.models import Config
+from users.models import User
+
+
+@pytest.fixture
+def config_system():
+    # TODO: Good enough for now, but a better Config mocking system is needed
+    result = Config.load_system()
+    with mock.patch("core.models.Config.load_system", return_value=result):
+        yield result
```
manfre, line 13: I intentionally avoided the start of a `core.test` or similar plumbing for now. This fixture is good enough for now, but I think we'll want something more along lines of Django's `[modify|override]_settings` classes.

andrewgodwin: Yeah, this definitely needs something like that.

andrewgodwin (approved):

PR #29, opened 2022-11-22
Content warning text

manfre:
* Added configuration option to change label and placeholder of "Content Warning".
* Created place for project wide fixtures. We may get to the point of needing a dumpdata fixture with a few different types of users and posts as we scale out the tests.


activities/views/timelines.py
```diff
@@ -31,6 +32,16 @@ class form_class(forms.Form):
             ),
         )
 
+        def __init__(self, *args, **kwargs):
+            super().__init__(*args, **kwargs)
+            if Config.system.content_warning_text:
```
andrewgodwin, line 37: You shouldn't need the `if` here - the default value should always be populated

users/views/admin/settings.py
```diff
@@ -37,6 +37,10 @@ class BasicSettings(AdminSettingsPage):
             "title": "Maximum Post Length",
             "help_text": "The maximum number of characters allowed per post",
         },
+        "content_warning_text": {
+            "title": "Content Warning Placeholder Text",
```
andrewgodwin, line 41: Let's call it "Content Warning Feature Name", and the help_text should be "What the feature that lets users provide post summaries is called"

takahe/test/fixtures.py
```diff
@@ -0,0 +1,43 @@
+import mock
+import pytest
+from django.conf import settings
+
+from core.models import Config
+from users.models import Domain, Identity, User
+
+
+@pytest.fixture(autouse=True)
```
andrewgodwin, line 9: I have done fixtures for most of this already on `main`

andrewgodwin (commented):
OK, so a few things:
* There's a Compose page that needs the same form changes done to it
* I've totally refactored the tests on `main` to all be in a top-level tests directory and not need the `__init__.py` files, and also done fixtures for a lot of what you have here

Sorry about the clash, but things are moving fast right now!

manfre:
I've applied the requested changes and updated for the (better) top level `tests` folder. pytest and its fixtures lends itself much nicer to mocking out config and I probably don't need the `(modify|override)_settings` helpers. A fixture to for the `Config.load_system` patching might be worthwhile for when tests want to inject the `config_system` fixture in to any view calls.

> Sorry about the clash, but things are moving fast right now!

No need to apologize, I know the risks ;)

core/models/config.py
```diff
@@ -55,6 +56,15 @@ class Meta:
 
     system: ClassVar["Config.ConfigOptions"]  # type: ignore
 
+    @classmethod
+    def lazy_system_value(cls, key: str):
```
andrewgodwin, line 60: Nice!

andrewgodwin:
Looks good - thanks!

PR #38, opened 2022-11-23
StatorRunner will refresh Config.system on each schedule_interval

manfre:
* Fixes a bug with `System.load_user` using wrong options_class.
* Aysnc `Config.aload_*`
* `StatorRunner` will freshen up `Config.system` on each scheduled interval


andrewgodwin (approved):

andrewgodwin:
Appreciate the naming of the new methods here

PR #39, opened 2022-11-23
Allow authenticated users to remote search/fetch by exact identity

manfre:


manfre:
I'm horrible about adding `Signed-off`. Let this comment serve as a declaration that unless I say otherwise at a point in the future, all of my commits to this repo have an implicit `Signed-off` by me.

andrewgodwin:
Don't worry about signed-off too much anyway, I don't really worry about it here

PR #40, opened 2022-11-23
Code dedupe Webfinger and fix SystemActor inbox url

manfre:
* SystemActor inbox url had erroneous `/`
* Fixed default `MAIN_DOMAIN` in development settings
* Make SystemActor quack more like an Identity for simplified WebFinger view

### DIFF
--- a/users/models/system_actor.py
+++ b/users/models/system_actor.py
@@ -22,6 +22,9 @@ class SystemActor:
         self.profile_uri = f"https://{settings.MAIN_DOMAIN}/about/"
         self.username = "__system__"
 
+    def absolute_profile_uri(self):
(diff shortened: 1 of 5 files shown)


users/models/system_actor.py
```diff
@@ -39,7 +42,7 @@ def to_ap(self):
         return {
             "id": self.actor_uri,
             "type": "Application",
-            "inbox": self.actor_uri + "/inbox/",
+            "inbox": self.actor_uri + "inbox/",
```
andrewgodwin, line 45: This reminds me that this inbox should probably go somewhere!

andrewgodwin (approved):

Issue #32, opened 2022-11-22
[help!] [docker] trying to rebuild my image


whatSocks:
Getting stuck trying to rebuild my image, I have tried many combinations of docker build, each failing in their own different way. Here's one:

➜  takahe git:(main) ✗ docker compose -f docker/docker-compose.yml up --build
[+] Building 0.4s (9/11)
 => [internal] load build definition from Dockerfile                                                                                                                                      0.0s
 => => transferring dockerfile: 592B                                                                                                                                                      0.0s
 => [internal] load .dockerignore                                                                                                                                                         0.0s
 => => transferring context: 2B                                                                                                                                                           0.0s
 => [internal] load metadata for docker.io/library/python:3.11.0-slim-buster                                                                                                              0.3s
 => [internal] load build context                                                                                                                                                         0.0s
 => => transferring context: 630B                                                                                                                                                         0.0s
 => [1/7] FROM docker.io/library/python:3.11.0-slim-buster@sha256:e3168357ada3287cda4c79ecae7f5103cb4d071d9b610a1d9380ea4b7cfc981f                                                        0.0s
 => CACHED [2/7] RUN apt-get update && apt-get -y install libpq-dev python3-dev build-essential                                                                                           0.0s
 => CACHED [3/7] COPY . /takahe                                                                                                                                                           0.0s
 => CACHED [4/7] WORKDIR /takahe                                                                                                                                                          0.0s
 => ERROR [5/7] COPY requirements.txt requirements.txt

andrewgodwin:
To build an image, the best way is `make image` from the root directory - I don't think docker-compose understands how.

PR #41, opened 2022-11-23
fix: use repository root as build context in `docker-compose.yml`

shuuji3:
Hello, I like this project! It has a nice clean and beautiful look and uses the latest Python async capability for scalability. 🙂 

When I tried `make compose_up`, I encountered the same error as (#17 and #32) where `requirements.txt` is missing. The error can be avoided if I run `make image` before running `make compose_up` but it seems that it doesn't solve the issue that `docker-compose build` fails after modifying some codes and rebuilding the container.

This PR fixes the issue where `docker-compose` fails to build `takahe:latest` container. The cause of the issue was the build context for docker-compose is different from the one of `docker build`. Currently, `make image` uses the repository root as a build context while `make compose_up` will use `docker/` directory as a build context. That's why docker-compose couldn't find the required files during building.

Documentation on docker website ([Compose file build reference | Docker Documentation](https://docs.docker.com/compose/compose-file/build/)) was helpful in understanding how to specify the correct context and `Dockerfile` path.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -2,7 +2,10 @@ version: "3.4"
 
 x-takahe-common:
   &takahe-common
-    build: .
+    build:
+      context: ..
+      dockerfile: ./docker/Dockerfile
+
     image: takahe:latest
     environment:
       DJANGO_SETTINGS_MODULE: takahe.settings.development


andrewgodwin (approved):

shuuji3:
Thanks for providing the docker-compose file anyway. It's handy for me to try the application on the local machine.

PR #48, opened 2022-11-25
Switch to Python 3.11 in tests to match docker images.

TkTech:
Also, ignore pycharm IDE files in .gitignore.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,10 +16,10 @@ jobs:
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
     steps:
       - uses: actions/checkout@v3
-      - name: Set up Python 3.10
+      - name: Set up Python 3.11
         uses: actions/setup-python@v4
         with:
-          python-version: "3.10"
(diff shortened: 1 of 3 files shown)


andrewgodwin:
And now you see why I did this (if you look at why the test build is failing)!

If you can get Black to format for 3.10 and the tests to run on 3.11 at the same time, I'll take _that_ PR :)

TkTech:
Interesting that the tests run fine in my fork, https://github.com/TkTech/takahe/actions/runs/3545407515/jobs/5953528969. Wonder if it's the cache.

Why not target 3.11 with black? The version you're using seems to support it.

andrewgodwin:
Because we're aiming for Python 3.10 as the low end of the compatibility range, which means that has to be the target format.

It also works fine on my system as I have Python 3.9, 3.10 and 3.11 all available, but that's not the case on the CI.

TkTech:
This seems like the simple solution. It runs both 3.10 and 3.11 tests and runs black against both.

andrewgodwin:
Lovely, thanks!

PR #49, opened 2022-11-25
Simplify settings

TkTech:
This is here for discussion, seems like you have a way you like to do settings and I wanted to offer a counterpoint.

This uses Pydantic for better settings validation, since pydantic was already a dependency.

Pros:

- Better error messages
- .venv file support (just make a .venv in your local folder, no need to rebuild docker images)
- No repetition with prefixes
- Configure database, email, storage with DSNs

Email is configured like so:

- `TAKAHE_EMAIL_BACKEND: "console://console"`
- `TAKAHE_EMAIL_BACKEND: "smtp://apikey:<api_key>@smtp.sendgrid.net:587?tls=true"`
- `TAKAHE_EMAIL_BACKEND: "smtp://apikey:<api_key>@smtp.sendgrid.net:587?ssl=true"`

Storage is configured like so:

- `TAKAHE_STORAGE_BACKEND: "s3://<key_id>:<key secret>@s3.us-east-2.amazonaws.com/<bucket name>`
- `TAKAHE_STORAGE_BACKEND: "gcs://google/<bucket name>`

Unfortunately you have to specify the whole S3 endpoint URL. However, I believe this is worthwhile because it enables using alternatives like minio (what I use) and S3-compatible endpoints like rackspace.

Database is configured like so:

- `TAKAHE_DATABASE_URL: "postgres://postgres:postgrese@db/takahe"`


takahe/settings.py
```diff
@@ -0,0 +1,242 @@
...
+    class Config:
+        env_prefix = "TAKAHE_"
+        env_file = str(BASE_DIR / ".env")
+        env_file_encoding = "utf-8"
+        # Case sensitivity doesn't work on Windows, so might as well be
+        # consistent from the get-go.
+        case_sensitive = False
+
+
+CONFIG = Settings()
```
andrewgodwin, line 77: The internal database-driven settings system is already called Config, so this will need a different name

manage.py
```diff
@@ -5,7 +5,7 @@
 
 # List of settings files that should guard against running certain commands
 GUARDED_ENVIRONMENTS = [
-    "production",
+    "prod",
```
andrewgodwin, line 8: I would like to keep this as "production" (and "development", etc.)

takahe/settings.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+class Settings(BaseSettings):
+    """
+    Pydantic-powered settings, to provide consistent error messages, strong
+    typing, consistent prefixes, .venv support, etc.
+    """
+
+    #: The default database.
+    DATABASE_URL: PostgresDsn
```
andrewgodwin, line 31: I would like the standard `PG*` environment variables to _also_ work, as they're meant to be common across software.

docker/docker-compose.yml
```diff
@@ -12,11 +12,15 @@ x-takahe-common:
...
-      PGPASSWORD: insecure_password
+      TAKAHE_DATABASE_URL: "postgres://postgres:insecure_password@db/takahe"
+      TAKAHE_DEBUG: true
+      TAKAHE_SECRET_KEY: "insecure_secret"
+      TAKAHE_CSRF_TRUSTED_ORIGINS: '["http://127.0.0.1:8000", "https://127.0.0.1:8000"]'
+      TAKAHE_USE_PROXY_HEADERS: true
+      TAKAHE_EMAIL_BACKEND: "console://console"
+      TAKAHE_MAIN_DOMAIN: "example.com"
+      TAKAHE_ENVIRONMENT: "dev"
+      DJANGO_SETTINGS_MODULE: "takahe.settings"
```
andrewgodwin, line 23: I would still like the ability to have a premade development settings rather than what you've done here, as I do not use docker-compose. At minimum, that means checking an env file in that matches these.

andrewgodwin (commented):
So I actually use Pydantic settings elsewhere and I do generally like them (hell, it's used for the internal Config system in a weird way), but there's a few things here I'd like to change first.

TkTech:
I believe I've addressed all your comments. There's an example .venv.dev, and validation to ensure there's either a `database_url` or a `PGHOST` set. Now, if you decide to merge the sqlite support, I'd suggest changing the default database to just be sqlite instead. With that, we can tweak the defaults a little so that 0 setup is required to get started, just creating the admin user.

andrewgodwin (approved):

andrewgodwin:
Thanks, I'll bring this in and then make the other few tweaks I want locally rather than having you do them!

PR #50, opened 2022-11-25
Test against sqlite & postgres

TkTech:
This is built on #48, #49 (so ignore everything but the last commit). It tests against both postgres and sqlite. It adds a (currently unused) boolean settings called `SEARCH` which can be used to gate full text search and other features that require a db that supports needed features.

From HN (and myself!) there's a lot of demand for sqlite. Since the service currently seems to work fine with sqlite, might as well test against it until it doesn't.


andrewgodwin:
I will have to ponder this one (and wait for the other PRs to merge before I can see a clean diff)!

andrewgodwin:
Feel like rebasing this given all the changes in the last few days?

Issue #27, opened 2022-11-20
Local-Only Post Visibility


whatSocks:
<img width="616" alt="Screen Shot 2022-11-20 at 5 21 23 PM" src="https://user-images.githubusercontent.com/2731744/202929432-e128c467-5311-47d1-9dc3-8118da7c66ab.png">

Learning from hometown the fork of Mastodon, local-only (ie: non-federated) posting can be a blessing. It can also be nice for testing little things (like boosts not ... boosting) in a production environment without broadcasting them to the entire fediverse. 

Check out the Hometown motivation here:

https://github.com/hometown-fork/hometown/wiki/Local-only-posting

The glitch-soc documentation here:

https://glitch-soc.github.io/docs/features/local-only-toots/

andrewgodwin:
Now this is a lovely idea I had not seen. Agreed!

PR #51, opened 2022-11-25
Local-only posting

manfre:
Initial effort for adding Local-Only support (#27)

@andrewgodwin wanted to get your generic feedback on this before trudging along too far. Does controlling the local vs federated entirely in fan_out and `to_ap()` work with your vision? ~~I don't like that the no-op fanouts generate a lot of chatter for non-local identities, but I haven't seen a better way yet.~~ (`Post.aget_targets()` filters out the potential no-op chatter)

**Notes:**
* There were unrelated updates in the migrations

### DIFF
--- a/activities/views/posts.py
+++ b/activities/views/posts.py
@@ -208,9 +208,8 @@ class Compose(FormView):
         ] = self.request.identity.config_identity.default_post_visibility
         if self.reply_to:
             initial["reply_to"] = self.reply_to.pk
+            initial["visibility"] = self.reply_to.visibility
             initial["text"] = f"@{self.reply_to.author.handle} "
-            if self.reply_to.visibility == Post.Visibilities.local_only:
-                initial["visibility"] = Post.Visibilities.local_only
         return initial
 
     def form_valid(self, form):
--- a/activities/views/posts.py
+++ b/activities/views/posts.py
@@ -172,6 +172,7 @@ class Compose(FormView):
         visibility = forms.ChoiceField(
             choices=[
                 (Post.Visibilities.public, "Public"),
+                (Post.Visibilities.local_only, "Local Only"),
                 (Post.Visibilities.unlisted, "Unlisted"),
                 (Post.Visibilities.followers, "Followers & Mentioned Only"),
                 (Post.Visibilities.mentioned, "Mentioned Only"),
@@ -207,8 +208,9 @@ class Compose(FormView):
         ] = self.request.identity.config_identity.default_post_visibility
         if self.reply_to:
             initial["reply_to"] = self.reply_to.pk
-            initial["visibility"] = Post.Visibilities.unlisted
             initial["text"] = f"@{self.reply_to.author.handle} "
+            if self.reply_to.visibility == Post.Visibilities.local_only:
+                initial["visibility"] = Post.Visibilities.local_only
         return initial
 
     def form_valid(self, form):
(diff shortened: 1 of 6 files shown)


activities/views/posts.py
```diff
@@ -200,12 +211,21 @@ def clean_text(self):
...
+        def __init__(self, *args, **kwargs):
+            super().__init__(*args, **kwargs)
+            initial = kwargs.get("initial")
+            if initial.get("reply_to"):
+                # Maintain local-ness of threads
+                self.fields["local_only"].disabled = True
+
     def get_initial(self):
         initial = super().get_initial()
+        initial["local_only"] = self.request.identity.config_identity.local_only_default
```
manfre, line 223: I don't like the the name of `config_identity` specifically because of the redundant `identity`, but naming it `config` is too close to typo realm of the reverse `configs` for my liking.

users/models/identity.py
```diff
@@ -429,3 +431,17 @@ def generate_keypair(self):
...
+    def config_identity(self) -> Config.IdentityOptions:
+        return Config.load_identity(self)
+
+    def lazy_config_value(self, key: str):
+        """
+        Lazily load a config value for this Identity
+        """
+        if key not in Config.IdentityOptions.__fields__:
+            raise KeyError(f"Undefined IdentityOption for {key}")
+        return lazy(lambda: getattr(self.config_identity, key))
```
manfre, line 447: I'm not sure if this lazy method makes sense. The tight coupling to an Identity means we probably can't use it during Form setup like we've done with Config.system values.

andrewgodwin:
I'd ideally like this done as a `Visibility` option rather than a separate boolean - you don't need local-only combined with, say, mentions only. It would just sit between "public" and "unlisted".

activities/migrations/0002_alter_post_visibility.py
```diff
@@ -0,0 +1,27 @@
...
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="post",
+            name="visibility",
+            field=models.IntegerField(
+                choices=[
+                    (0, "Public"),
+                    (4, "Local Only"),
```
andrewgodwin, line 19: Since CHOICES doesn't affect the database, patch this into the previous migration instead if you could!

activities/models/fan_out.py
```diff
@@ -44,13 +46,16 @@ async def handle_new(cls, instance: "FanOut"):
                         identity=fan_out.identity,
                         post=post,
                     )
-            else:
+            elif not post.visibility != Post.Visibilities.local_only:
```
andrewgodwin, line 49: Hmm, can we do this in the PostStates when we make the fan-outs, by removing non-local identities there?

manfre: I think the filtering in `Post.aget_targets()` makes the checks here redundant. I was being a bit more explicit cautious while still wrapping my brain around the state logic.

activities/views/posts.py
```diff
@@ -207,8 +208,9 @@ def get_initial(self):
         ] = self.request.identity.config_identity.default_post_visibility
         if self.reply_to:
             initial["reply_to"] = self.reply_to.pk
-            initial["visibility"] = Post.Visibilities.unlisted
```
manfre, line 210: Should reply_to visibility ignore `default_post_visibility` and force `unlisted` when dealing with federating posts?

andrewgodwin: It should actually default to the visibility of the post you're replying to if that's unlisted/followers only/mentions only, or unlisted if that's public/local (we don't want to default it to unlisted if the original message is mentions-only!).

andrewgodwin (approved):

andrewgodwin:
Thanks!

PR #52, opened 2022-11-26
Add default post visibility

manfre:
* Adds `default_post_visibility` Identity setting.
* Splits `users.views.settings.__init__` to be more consistent of one page/view per file.


andrewgodwin (approved):

andrewgodwin:
Thanks!

PR #54, opened 2022-11-26
Don't error on admin panel if user is anonymous

TkTech:
Right now going to /admin with an AnonymousUser just errors instead of redirecting to /login.


TkTech:
Whoops, based off the wrong branch. Will fix this PR after breakfast.

TkTech:
Beat me to it!

PR #56, opened 2022-11-26
Post editing

manfre:
Initial pass at post editing. 

* I changed the fan out to use match/case for better readability
* compose/edit form is implemented
* ~~The state transitions and fanout are commented out due to me not understanding how to use it properly.~~ I think I get it now. I like it.
* Self review comments to follow that point out assumptions and questions

### DIFF
--- a/activities/models/fan_out.py
+++ b/activities/models/fan_out.py
@@ -23,8 +23,8 @@ class FanOutStates(StateGraph):
         fan_out = await instance.afetch_full()
 
         match (fan_out.type, fan_out.identity.local):
-            # Handle creating local posts
-            case (FanOut.Types.post, LOCAL_IDENTITY):
+            # Handle creating/updating local posts
+            case (FanOut.Types.post | FanOut.Types.post_edited, LOCAL_IDENTITY):
                 post = await fan_out.subject_post.afetch_full()
                 # Make a timeline event directly
                 # If it's a reply, we only add it if we follow at least one
@@ -49,7 +49,7 @@ class FanOutStates(StateGraph):
                         post=post,
                     )
 
-            # Handle sending remote posts creation
+            # Handle sending remote posts create
             case (FanOut.Types.post, REMOTE_IDENTITY):
                 post = await fan_out.subject_post.afetch_full()
                 # Sign it and send it
@@ -57,6 +57,16 @@ class FanOutStates(StateGraph):
                     method="post",
                     uri=fan_out.identity.inbox_uri,
                     body=canonicalise(post.to_create_ap()),
+                )
+
+            # Handle sending remote posts update
+            case (FanOut.Types.post_edited, REMOTE_IDENTITY):
+                post = await fan_out.subject_post.afetch_full()
+                # Sign it and send it
+                await post.author.signed_request(
+                    method="post",
+                    uri=fan_out.identity.inbox_uri,
+                    body=canonicalise(post.to_update_ap()),
                 )
 
             # Handle deleting local posts
--- a/activities/models/fan_out.py
+++ b/activities/models/fan_out.py
@@ -17,11 +17,15 @@ class FanOutStates(StateGraph):
         """
         Sends the fan-out to the right inbox.
         """
+        LOCAL_IDENTITY = True
+        REMOTE_IDENTITY = False
+
         fan_out = await instance.afetch_full()
-        # Handle Posts
-        if fan_out.type == FanOut.Types.post:
-            post = await fan_out.subject_post.afetch_full()
-            if fan_out.identity.local:
+
+        match (fan_out.type, fan_out.identity.local):
+            # Handle creating local posts
+            case (FanOut.Types.post, LOCAL_IDENTITY):
+                post = await fan_out.subject_post.afetch_full()
                 # Make a timeline event directly
                 # If it's a reply, we only add it if we follow at least one
                 # of the people mentioned.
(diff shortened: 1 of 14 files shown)


activities/models/fan_out.py
```diff
@@ -17,11 +17,15 @@ async def handle_new(cls, instance: "FanOut"):
...
         fan_out = await instance.afetch_full()
-        # Handle Posts
-        if fan_out.type == FanOut.Types.post:
-            post = await fan_out.subject_post.afetch_full()
-            if fan_out.identity.local:
+
+        match (fan_out.type, fan_out.identity.local):
+            # Handle creating/updating local posts
+            case (FanOut.Types.post | FanOut.Types.post_edited, LOCAL_IDENTITY):
+                post = await fan_out.subject_post.afetch_full()
```
manfre, line 28: This treats new and updates identically without caring to clean up target differences between original and updated. Should it?

activities/models/fan_out.py
```diff
@@ -44,63 +48,91 @@ async def handle_new(cls, instance: "FanOut"):
...
                     uri=fan_out.identity.inbox_uri,
                     body=canonicalise(interaction.to_ap()),
                 )
-        # Handle undoing boosts/likes
-        elif fan_out.type == FanOut.Types.undo_interaction:
-            interaction = await fan_out.subject_post_interaction.afetch_full()
-            if fan_out.identity.local:
+
+            # Handle undoing local boosts/likes
+            case (FanOut.Types.undo_interaction, LOCAL_IDENTITY):  # noqa:F841
```
manfre, line 112: `noqa` due to flake8 having issues with only some of the case statements.

activities/views/posts.py
```diff
@@ -203,28 +207,53 @@ def clean_text(self):
...
+        post_id = form.cleaned_data.get("id")
+        if post_id:
+            post = get_object_or_404(self.identity.posts, pk=post_id)
+            post.content = form.cleaned_data["text"]
+            post.summary = form.cleaned_data.get("content_warning")
+            post.visibility = form.cleaned_data["visibility"]
+            post.save()
+
+            # Should there be a timeline event for edits?
+            # E.g. "@user edited #123"
```
manfre, line 239: Should there be a timeline event for edits? If not, do we want to indicate on the post view that it has been edited?

manfre (commented):
Self review complete

activities/models/fan_out.py
andrewgodwin: That's kind of how the rest of the code treats it, so I think we're fine. The only thing I'd care about is marking that a post has been edited, which we can retrieve from the post model directly.

activities/views/posts.py
andrewgodwin: Timeline event no, marked on post yes. The `edited` field on the Post object is intended to be this flag.

activities/models/post.py
```diff
@@ -141,6 +164,7 @@ class urls(urlman.Urls):
         action_unboost = "{view}unboost/"
         action_delete = "{view}delete/"
         action_reply = "/compose/?reply_to={self.id}"
+        action_edit = "/compose/{self.id}/edit/"
```
andrewgodwin, line 167: This URL should probably either be `{view}edit/` or `/compose/?edit={self.id}`. Any preference?

activities/models/post.py
```diff
@@ -336,6 +360,20 @@ def to_create_ap(self):
...
+        Returns the AP JSON to update this object
+        """
+        object = self.to_ap()
+        return {
+            "to": object["to"],
+            "cc": object.get("cc", []),
+            "type": "Update",
+            "id": self.object_uri + "#update",
+            "actor": self.author.actor_uri,
+            "object": object,
```
andrewgodwin, line 374: We also need to update `to_ap` to transform the local `edited` field into the AP `updated` field if it's present.

andrewgodwin (commented):
Didn't read through the async test stuff yet, but here's a few small changes on the feature side. As for where to show the icon that a post has been edited - let's stick a `<small>(Edited 3d ago)</small>` at the end of the post text for now and we'll style it so it's dimmer and smaller.

activities/views/posts.py
```diff
@@ -203,28 +207,56 @@ def clean_text(self):
...
+            )
+            # Add their own timeline event for immediate visibility
+            TimelineEvent.add_post(self.request.identity, post)
         return redirect("/")
 
-    def dispatch(self, request, *args, **kwargs):
+    def dispatch(self, request, post_id=None, *args, **kwargs):
+        self.post_obj = None
+        if post_id:
+            self.post_obj = get_object_or_404(request.identity.posts, pk=post_id)
```
andrewgodwin, line 258: This needs to check that the post author is the same as the current request identity.

activities/models/post.py
manfre: Switching to `{view}edit/`

activities/views/posts.py
manfre: `request.identity.posts` already ensures the request user only sees their own Post. Or are you asking for an explicit HTTP 403, instead of the 404? Either way, I'll add a test to ensure behavior.

manfre: I think I understand the confusion. `Delete` was using `self.identity` (based upon URL) and `Compose` was using `request.identity`. I've updated `Compose` and `Delete` to return a 403 (need styled error pages) with a explicit check of handle against `request.identity`.

tests/activities/models/test_post.py
```diff
@@ -112,3 +115,45 @@ def test_linkify_mentions_local(identity, remote_identity):
         local=True,
     )
     assert post.safe_content_local() == "<p>@test@example.com, welcome!</p>"
+
+
+async def stator_process_tasks(stator):
```
andrewgodwin, line 120: I think I might just move this into the stator runner, we do need something like this!

andrewgodwin (approved):

PR #57, opened 2022-11-26
Docs

TkTech:
- Ensure documentation builds as part of the tests.
- On release, build and deploy documentation.

Right now you're using readthedocs, but IMO I prefer to use a minimum of tools, so I tend to use this pattern. If you want to use this, create a CNAME that points docs.jointakahe.org to jointakahe.github.io.


andrewgodwin:
I like using readthedocs due to their support for multiple versions, so I'd take this if it's just the tests part.

PR #61, opened 2022-11-27
Specify settings .env file with TAKAHE_ENV_FILE

manfre:
Allows specifying the settings file using `TAKAHE_ENV_FILE=whatever.env` with `.env` as the default.

```
TAKAHE_ENV_FILE=test.env pytest
TAKAHE_ENV_FILE=local.env python manage.py runserver
```

### DIFF
--- a/takahe/settings.py
+++ b/takahe/settings.py
@@ -1,3 +1,4 @@
+import os
 import secrets
 import sys
 import urllib.parse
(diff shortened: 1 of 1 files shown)


andrewgodwin:
If you rebase the tests should pass and I can merge this.

FWIW, I landed a commit to automatically use `test.env` when pytest is running, so that should solve your actual problem here with less typing.

manfre:
Rebased

manfre:
I tweaked the PR because the explicit `Settings(env_file=...)` overrides for every manage.py because `pytest` is in my venv.

andrewgodwin:
Yeah, this is better than my hack from last night!

PR #62, opened 2022-11-27
Heroku buildpack support.

TkTech:
This adds the necessary metadata for Heroku, Dokku, and the other platforms that support Heroku buildpacks. Currently, it's for advanced usage only (there's no app.json) and there's a quick draft Dokku guide here: https://tkte.ch/articles/2022/11/26/Takahe-and-Dokku.html


TkTech:
This _should_ also work for fly.io, but is untested.

andrewgodwin:
I do wish all these hosting companies wouldn't mess up the root directory, but I guess there's not much we can do about that!

TkTech:
Agreed! At least a bunch of them have semi-standardized, better than having ten files for each!

PR #68, opened 2022-11-28
Test support for sqlite.

TkTech:


andrewgodwin:
Looks like I merged in the wrong order and this is conflicting again - one last rebase?

TkTech:
Updated

andrewgodwin:
Lovely. Curious how this ends up working with hashtag search, as that's a JSON operator!

TkTech:
Should work fine, just needs the JSON extension to be loaded into sqlite which is the default on most (all?) distro distributions and most binary python distributions.

rcarmo:
Actually, it doesn't on a vanilla installation of Ubuntu 22.04 (arm64). I had to [move to PostgreSQL](https://taoofmac.com/space/blog/2022/12/21/0900) to make the stator happy.

I stumbled upon this by chance, am going to investigate further.

PR #70, opened 2022-11-28
Relax S3 URI parsing

tabletcorry:
Create a new URL validator for `MEDIA_BACKEND` to allow for S3 specific quirks (mainly, no host). Also modify the S3 parsing code to allow for lack of username/password/host.

### DIFF
--- a/takahe/settings.py
+++ b/takahe/settings.py
@@ -13,6 +13,11 @@ from sentry_sdk.integrations.django import DjangoIntegration
 BASE_DIR = Path(__file__).resolve().parent.parent
 
 
+class MediaBackendUrl(AnyUrl):
+    host_required = False
+    allowed_schemes = {'s3', 'gcs', 'local'}
+
+
 def as_bool(v: Optional[Union[str, List[str]]]):
     if v is None:
         return False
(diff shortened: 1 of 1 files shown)


tabletcorry:
Hat tip to @TkTech for https://github.com/jointakahe/takahe/pull/68/files#diff-9a840a2c4ff3edf62750c7ca5a9b21edf7aecc8bfa2785202f7dbd19cb3be780R16

takahe/settings.py
```diff
@@ -299,10 +304,12 @@ class Config:
     elif parsed.scheme == "s3":
         DEFAULT_FILE_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"
         AWS_STORAGE_BUCKET_NAME = parsed.path.lstrip("/")
```
manfre, line 306: Not sure if in scope for this PR, but we can also extract a value for `AWS_LOCATION` from the path with `split("/", 1)` .

Something along the lines of:
```
s3_parts = parsed.path.lstrip("/").split("/", 1)
AWS_STORAGE_BUCKET_NAME = s3_parts[0]
if len(s3_parts) == 2:
    AWS_LOCATION = s3_parts[1]
```

tabletcorry: Probably out of scope until we figure out path storage in a shared bucket?

Or, more possibly, pushing images through a CDN front.

andrewgodwin (approved):

tabletcorry:
@andrewgodwin Cleared up the linting issue, should be good to go

PR #71, opened 2022-11-28
Hashtags

manfre:
PR provides:
* Hashtags are parsed from post content
* Hashtag record is stored for the purpose of 
  * collecting stats
  * moderation (`public=False`) and Config setting to hide unreviewed
  * providing a human friendly display name `#BecauseTagsLikeThisAreImpossibleToReadWithoutCamelCase`
* Stats are generated (based upon what would show on the hashtag timeline)
* Local hashtag timeline page exists at `/tags/<hashtag>/`
* Trending tags page exists at `/explore/tags/`. It currently generates off the total usage count.
* Django Admin actions to re-parse existing posts to extract hashtags
* Django Admin action to manual force Hashtags to be outdated to trigger stats recalculation
* Hashtag search

To do or defer for another PR:
* I'm horrible at UI. Help appreciated.
* ~~`Hashtag.stats` needs to be computed periodically and eventually expired to avoid unbounded growth. Maybe `Hashtag.perform_transition("outdated")` whenever we see a tag is good enough and provides a natural falloff.~~
* Create template tag that leverages `Hashtag.usage_days()` to show a chart or something more trending that `Post.stats["total"]`
* Convert existing `Hashtag` to an alias of another Hashtag.


activities/models/hashtag.py
```diff
@@ -0,0 +1,145 @@
...
+from asgiref.sync import sync_to_async
+from django.db import models
+from django.utils import timezone
+
+from stator.models import State, StateField, StateGraph, StatorModel
+
+
+class HashtagStates(StateGraph):
+    outdated = State(try_interval=300)
+    updated = State(externally_progressed=True)
```
andrewgodwin, line 15: If you want periodic recomputation, I have a pattern for that on Identity now

activities/models/hashtag.py
```diff
@@ -0,0 +1,145 @@
...
+class Hashtag(StatorModel):
+
+    # Normalized hashtag without the '#'
+    hashtag = models.SlugField(primary_key=True, max_length=100)
+
+    # Friendly display override
+    name_override = models.CharField(max_length=100, null=True, blank=True)
+
+    # Should this be shown in the public UI?
+    public = models.BooleanField(default=True)
```
andrewgodwin, line 70: Is the intent for this to be the moderation of hashtags into, say, trending? If so, we may want to have this be three way (nullable BooleanField) with the default of null meaning "unreviewed, don't show"

activities/models/post.py
```diff
@@ -35,18 +36,33 @@ class PostStates(StateGraph):
     edited_fanned_out.transitions_to(deleted)
 
     @classmethod
-    async def handle_new(cls, instance: "Post"):
-        """
-        Creates all needed fan-out objects for a new Post.
-        """
-        post = await instance.afetch_full()
+    async def ensure_hashtags(cls, post: "Post") -> None:
```
andrewgodwin, line 39: This should probably be a Post method, not on the states.

andrewgodwin (commented):
Looking pretty good so far! I didn't see any code that creates links for hashtags in posts to their relevant tag browse page - did I miss it, or not in yet?

activities/models/hashtag.py
manfre: I went with binary state to treat moderation as a reactive correction with the ability to preempt spammy hashtags by prepopulating. I can change to trinary if you'd prefer.

manfre: Now trinary with a `Config.system` to specify if unreviewed is treated as public True or False

andrewgodwin (approved):

PR #72, opened 2022-11-28
Fix case typo in email settings

tabletcorry:
The `L` in `EMAIL_HOST_USER` has the wrong case in the SMTP parser, which had me tearing my hair out for a while (if _either_ user or password aren't set, Django Email just skips auth completely, rather than returning any kind of sane error)


andrewgodwin (approved):

andrewgodwin:
Oof, how did we miss that! (by using the sendgrid shortcut on my end, at least). Thanks!

PR #73, opened 2022-11-28
docs: fix code block syntax

shuuji3:
This PR contains two changes:
- Just a trivial markup fix
- Add GitHub Actions workflow to build documentation 

### Markup fix

The markup fix is just adding a missing colon (`:`) in the Contributing page: [Contributing - Takahē documentation](https://docs.jointakahe.org/en/latest/contributing/#direct-installation)

**before**
Screenshot 2022-11-28 at 23 20 03

**after**
Screenshot 2022-11-28 at 23 19 51

### GitHub Actions workflow for docs

I also add a new workflow only targeting docs to ensure buildable and limiting the condition of `test.yml`. With this change, commits changing files under `docs/**` will trigger only `test-docs.yml` and can avoid triggering `test.yml`, which consumes a bit longer time.

### DIFF
--- a/.github/workflows/test-docs.yml
+++ b/.github/workflows/test-docs.yml
@@ -1,0 +1,29 @@
+name: Test Documentation Build
+
+on:
+  push:
+    paths:
+      - 'docs/**'
+  pull_request:
+    paths:
+      - 'docs/**'
+
+jobs:
+  test_docs:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.10"]
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: pip
+      - name: Install dependencies for docs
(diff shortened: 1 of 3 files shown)


.github/workflows/test-docs.yml
```diff
@@ -0,0 +1,29 @@
...
+  pull_request:
+    paths:
+      - 'docs/**'
+
+jobs:
+  test_docs:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.10"]
```
shuuji3, line 16: I remain only 3.10 to match the same version in https://github.com/jointakahe/takahe/blob/main/.readthedocs.yaml#L7.

TkTech:
The path exclusions should probably be removed - none of the files in docs/ may have changed, but it's possible for the documentation to reference the source and thus break when it changes.

shuuji3:
Ah, I missed that case. I'll remove the condition. `test-docs.yml` should be run always. Thanks for pointing that out. 🙂

andrewgodwin (approved):

andrewgodwin:
Thanks!

PR #74, opened 2022-11-28
Update phony targets

eriol:
Just update phony targets, although it will be extremely rare to have files with the name of the targets in the Makefile I think it's better to keep it updated.


andrewgodwin:
Thanks!

eriol:
Thanks for merging!

PR #78, opened 2022-11-29
unquote SMTP and AWS passwords

tabletcorry:
Passwords can contain characters that are illegal in URL identities. So use unquote to solve that.

WARNING: This means that a freak URL escape in a password will fail in the opposite way. But since this is a URL, everything _must_ be urlencoded for sanity.


andrewgodwin (approved):

andrewgodwin:
Yup, seems very sensible - thanks!

PR #82, opened 2022-11-30
Add copy to clipboard action to user's handle

### LINKED ISSUE #34: Copy-to-clipboard icon for easy manual remote follow of users and posts
As a stepping stone to remote follow, it would be nice to have a copy to clipboard icon next to user's username@domain on the public feed for unauthed users (for user search) and also on the post on the top right (for post search)

humrochagf:
Resolves #34

### DIFF
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -1050,3 +1050,19 @@ table.metadata td.name {
         display: none;
     }
 }
+
+/* Copy to clipboard action */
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Lovely! Nice use of hyperscript.

humrochagf:
Thanks 😊

PR #83, opened 2022-12-01
Tweak UI when no posts found for remote identity

aviflax:
I’ve found the current copy confusing when viewing a remote identity.


andrewgodwin:
Yup, nice addition. Hopefully we'll eventually add "retrieving old posts from the user's outbox" to our identity fetcher!

PR #86, opened 2022-12-01
Add a special release task to Procfile to run migrations before startup.

TkTech:
A task called release is treated differently from the rest. Both Heroku and dokku will run this _just once_ before the container starts up.

See:
- https://devcenter.heroku.com/articles/release-phase#specifying-release-phase-tasks
- https://dokku.com/docs/advanced-usage/deployment-tasks/#procfile-release-command


andrewgodwin:
Thanks!

PR #88, opened 2022-12-03
Direct installation contributing guide updates

pauloxnet:


docs/contributing.rst
```diff
@@ -82,32 +82,41 @@ Then, ``cd`` into that directory and create and activate a virtual environment
 
 Then install the development requirements::
 
-    pip install -r requirements-dev.txt
+    python3 -m pip install -r requirements-dev.txt
```
andrewgodwin, line 85: Curious, what aspect of your local system stops all the binaries being on the PATH?

pauloxnet: These changes are not specifically for me but for every potential contributors of the project. Using `pip` as module is a best practices I started following years ago and I saw a lot of other projects are using as Django itself
https://docs.djangoproject.com/en/dev/topics/install/#installing-an-official-release-with-pip

One of the best explanation of this best practices is the one you can read in the  Brett Cannon's blog I think:
https://snarky.ca/why-you-should-use-python-m-pip/

andrewgodwin:
Ah, fair enough - I do know it's best practice, I have just yet to see a developer with a fully-working Python environment and virtualenv where the PATH stuff doesn't work!

Issue #84, opened 2022-12-01
When logged in, searches for remote identities yield a result only if repeated


aviflax:
Here’s how it goes:

1. I open the search page
2. I enter an identity… uh… identifier? I’m not sure what these are called. For example `@avi@flaximus.net`
3. I click the _Search_ button
4. The page refreshes. The form field still has the user identifier I entered, and the form otherwise is blank. There are no results, no text saying “no results”, nothing.
5. I click the _Search_ button again
6. The page refreshes. This time below the form, there’s a header _People_ and below that, one search result is displayed; the person I was searching for.

I’m not even sure how I discovered this. I suppose there’s a good chance other folks won’t discover it and will just assume that searching for identities is just broken. Anyway, it would be better if the results would show up the first time I click _Search_. Failing that, we should at least display some copy explaining what’s going on, and that the user should try clicking _Search_ once again.

I’m currently running [c728933](https://github.com/jointakahe/takahe/commit/c72893372afe255689ff56a618b6c849974569a3) on a very wimpy server (see #80)

andrewgodwin:
My guess is that this is because the search page is doing a background fetch of the identity between the two calls; if you watch Mastodon's search feature, they just do it as part of the process and make the load take longer, which I think is probably reasonable (once we add some timeouts to our calls).

PR #92, opened 2022-12-04
Tweaks to search UX

aviflax:
including a band-aid for #84


andrewgodwin:
Thanks - this will help until I get a chance to see if I can make search do identity fetching inline!

PR #99, opened 2022-12-04
Add TAKAHE_DEFAULT_TIMEOUT with default of 5.0

manfre:
* Timeout of 5.0 was arbitrary (and generous) to avoid blocking until gunicorn cuts the connection
* Start of a tuning docs page.


andrewgodwin:
Nice, I was about to go do this today when you read my mind and got it done first!

PR #100, opened 2022-12-04
Remove trailing slashes from most resource URLs

aviflax:
Because:

* they seem probably unnecessary
* IMO they detract slightly from the aesthetics & usability of the URLs
* they’re not idiomatic for the Web, in general, as I understand such things
* they’re not [idiomatic][iwf], and seemingly incorrect, for [WebFinger]
  * in the spec the “well-known” URI suffix is just `webfinger`
* they may be non-idiomatic, and may be incorrect, for [ActivityStreams]
  * the examples I saw in the spec did not have trailing slashes

The tests pass, but I haven’t had a chance to run the app with these changes and “click around”.

[iwf]: https://docs.joinmastodon.org/spec/webfinger/
[WebFinger]: https://www.rfc-editor.org/rfc/rfc7033.html
[ActivityStreams]: https://www.w3.org/TR/activitystreams-core/


andrewgodwin:
I'm afraid I must reject this one - the Django standard is for trailing slashes everywhere, and it appears to work perfectly fine with all the ActivityPub servers I've tried it on. Webfinger is the only case where we have to omit it by convention, and so we do so.

Also, now there's live Takahē installations out in the wild, the URL format for post and actor URIs is locked in and cannot be changed, as otherwise everything would break.

PR #106, opened 2022-12-04
Optimized deb packages installed in Docker

pauloxnet:
I optimized the Deb packages installed in Docker, making the installation of the development libraries temporary for the installation of the Python requirements and leaving the one package needed to run psycopg permanent.

I have also removed the installation of unnecessary packages. Updating pip is an extra layer that should be addressed by updating the official Python image used as a base.

The final image is thus reduced by 50% for the same features.

#### Before
```bash
(takahe) paulox@net:~/Projects/takahe$ docker image ls
REPOSITORY        TAG                  IMAGE ID       CREATED              SIZE
takahe            latest               c9901e63c721   About a minute ago   980MB
postgres          15-alpine            9104349146d9   3 days ago           243MB
python            3.11.0-slim-buster   20d9cc821c50   2 weeks ago          121MB
```

#### After
```bash
(takahe) paulox@net:~/Projects/takahe$ docker image ls
REPOSITORY        TAG                  IMAGE ID       CREATED          SIZE
takahe            latest               28f8faab0ad9   12 seconds ago   528MB
postgres          15-alpine            9104349146d9   3 days ago       243MB
python            3.11.0-slim-buster   20d9cc821c50   2 weeks ago      121MB
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,12 +3,24 @@ ARG IMAGE_LABEL=3.11.0-slim-buster
 
 FROM ${IMAGE_HOST}:${IMAGE_LABEL}
 
-RUN apt-get update && apt-get -y install libpq-dev libxml2-dev libxslt1-dev zlib1g-dev python3-dev build-essential
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        libpq5 \
+    && rm -rf /var/lib/apt/lists/*
 
 COPY requirements.txt requirements.txt
 
-RUN pip3 install --upgrade pip \
-    && pip3 install --upgrade -r requirements.txt
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        gcc \
+        libc6-dev \
+        libpq-dev \
+    && python3 -m pip install --upgrade -r requirements.txt \
+    && apt-get purge -y --auto-remove \
+        gcc \
+        libc6-dev \
+        libpq-dev \
+    && rm -rf /var/lib/apt/lists/*
 
 COPY . /takahe
 


TkTech:
This may be a good candidate for switching to the alpine base image, which would drop the base image size from 50mb to 15mb.

pauloxnet:
Alpine images for Python it's a bad idea IMO
https://pythonspeed.com/articles/alpine-docker-python/

aviflax:
FWIW this LGTM, but it does make these steps a little harder to parse visually and thus possibly slightly more prone to maintenance errors. No big deal but I would think an alternative might be to use a multi-stage build; we could download and install all the deps in a build stage, install them into a specific target directory, and then in the final stage, copy them from the build stage. In that case, no cleanup is required, because whatever build tools were installed into the build stage are omitted from the final stage.

andrewgodwin:
I tried doing this as a multi-stage build based on the previous one of those, but doing it without having a virtualenv (because I want to use system packages for some Python stuff in future) proved very hard to get working.

pauloxnet:
A tried to modify the file as little as possible with maximum reduction of size without change base image and without made a radical change (like multistage).

But I think a multi stage approach can be a good idea, maybe in an another PR.

In the meantime this PR cut in half the size ( from ~1GB to ~500MB)

TkTech:
> Alpine images for Python it's a bad idea IMO https://pythonspeed.com/articles/alpine-docker-python/

This article was pretty disputed at the time, and most of its points are moot as of 2022, FYI.

pauloxnet:
I still don't think Alpine is not a good idea. But for sure I would propose that change in another PR.

manfre:
> I tried doing this as a multi-stage build based on the previous one of those, but doing it without having a virtualenv (because I want to use system packages for some Python stuff in future) proved very hard to get working.

I think the `python:*` images use a compiled python and not system one as default. Trying to leverage system packages while also using the tag specified version might be problematic.

pauloxnet:
I'll be happy to review other PRs with the various approaches I've read about here.

But in my opinion for the moment it would be useful to merge this PR since it removes the installation of totally useless packages (eg: python-dev) and at the end of the build process it leaves only libpq for Psycopg2 in the image.

P.S. when Django 4.2 will be released we will also have the backend for Psycopg v3 and we will be able to install the C compiled version without needing to install any compiler or development library.

andrewgodwin:
I still want to stick with a Debian base for now, out of personal preference for working with it if nothing else.

I'll get this one merged in though - thanks!

PR #107, opened 2022-12-04
Make max media upload size configurable

aviflax:
As some of us want to upload full-size photos and videos.

I’m pretty rusty with Django so sorry if this isn’t the idiomatic way to access the setting from a view. Please let me know if not, and I’ll fix this.

I hope to propose a follow-up change that’ll make the max image dimensions configurable. After that, someone should be able to configure their installation such that they’d be able to post, say, a 14 megabyte file of a 25 megapixel photo, and have it be uploaded, saved, and attached to a post, at the original fidelity. If so desired.

### DIFF
--- a/activities/views/compose.py
+++ b/activities/views/compose.py
@@ -1,4 +1,5 @@
 from django import forms
+from django.conf import settings
 from django.core.exceptions import PermissionDenied
 from django.shortcuts import get_object_or_404, redirect, render
 from django.utils.decorators import method_decorator
@@ -158,10 +159,12 @@ class ImageUpload(FormView):
 
         def clean_image(self):
             value = self.cleaned_data["image"]
-            if value.size > 1024 * 1024 * 10:
+            max_mb = settings.SETUP.POST_MEDIA_UPLOAD_MAX_MB
+            max_bytes = max_mb * 1024 * 1024
+            if value.size > max_bytes:
                 # Erase the file from our data to stop trying to show it again
                 self.files = {}
-                raise forms.ValidationError("File must be 10MB or less")
+                raise forms.ValidationError(f"File must be {max_mb}MB or less")
             return value
 
     def form_invalid(self, form):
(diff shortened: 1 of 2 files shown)


andrewgodwin:
It's worth noting why this is in place - Mastodon's image proxy won't support super large images, so we forcibly scale all uploaded media down so it's an acceptable size to give out to Mastodon so it can load them (2000x2000). That means increasing the uploaded media size will merely give the resizer more work to do and not keep the originals around.

If you also want to keep the originals, we will need to add a third FileField to PostAttachment that stores the original image, and then not send that over ActivityPub or client APIs but only use it when viewed locally. I opted not to do that yet because of the very small utility.

andrewgodwin:
Yes, that's a good point - this is more just a setting right now of "how much RAM did you give the web process?". 

With that in mind, can we rename the setting to `MEDIA_MAX_IMAGE_FILESIZE`? That way it lines up with the other media settings, and we can also implement an alternative one when we get to videos, as those need background processing anyway.

aviflax:
> Yes, that's a good point - this is more just a setting right now of "how much RAM did you give the web process?".

👍
 
> With that in mind, can we rename the setting to `MEDIA_MAX_IMAGE_FILESIZE`? That way it lines up with the other media settings, and we can also implement an alternative one when we get to videos, as those need background processing anyway.

Of course. How would you feel about adding the suffix `_MB` to make the units clear? I know the resulting var name is pretty long, but I find it really helpful when numeric configs indicate the units right in the name.

andrewgodwin:
Yup, agreed, let's stick the unit on there. Better to be explicit.

aviflax:
Done via 56ec2e0a55da372b62847505de2a07b19aadde80

aviflax:
apropos to the shelved-for-now notion of increasing image resizing dimensions, I found an ongoing and in-depth discussion of image resolutions here: mastodon/mastodon#20255

andrewgodwin:
Thanks. That issue you linked does seem rather split - there's definitely a "phone cameras will just burp out giant files" versus "artists want to preserve their work" split. Odd that Mastodon forces everything to low-quality JPEG, though, not webp, given the compatibility of the latter!

PR #109, opened 2022-12-05
Add TAKAHE_CACHE_DEFAULT to configure the default cache

manfre:
* Cache DSN configured with `TAKAHE_CACHE_DEFAULT`
* `ConfigLoadingMiddleware` will only reload `Config.system` periodically (or if needed)
* Decorated most views with `cache_page` equivalent that has awareness of `Config.system` and identities (when desired)
* Added `Tuning` settings page

### DIFF
--- a/activities/views/follows.py
+++ b/activities/views/follows.py
@@ -1,11 +1,15 @@
 from django.utils.decorators import method_decorator
 from django.views.generic import TemplateView
 
+from core.decorators import per_identity_cache_page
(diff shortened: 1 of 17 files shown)


manfre:
This PR is ready. The timeouts are somewhat arbitrary and it might make sense to add `Config.system` or `TAKAHE_` values to fine tune them.

andrewgodwin:
Thanks! I'll take it for a test drive on takahe.social later.

PR #110, opened 2022-12-05
Support deeper subdomains in domain validation

tomusher:
The regex used for domain validation doesn't support 4th-level or greater subdomains.

The [SO post](https://stackoverflow.com/a/26987741/23746) that had the original regex used here has an update with a more complete pattern, so I've updated the pattern here to match.


shuuji3:
This is a good catch! I'm not sure but I also found the `host_re` regex in Django's built-in `URLValidator` class: [django.core.validators | Django documentation | Django](https://docs.djangoproject.com/en/4.1/_modules/django/core/validators/#URLValidator). Maybe could we use this regex here? It seems to cover more comprehensive patterns.

andrewgodwin:
Oh yes, we should totally use the Django one if we can - it's been fixed over many years of bug reports.

tomusher:
I've updated this to use a new DomainValidator based on the regex from Django's URLValidator and added a few tests.

pauloxnet:
I think it might be better to reference the regex already present in the Django code rather than duplicate the regex with the risk that they soon become not aligned.

tomusher:
I felt like pulling undocumented attributes directly off `URLValidator` has the potential to break with Django upgrades, and subclassing `URLValidator` and trying to 'undo' all the bits we don't need feels fragile and unnecessarily complicated - but I'm happy to rework this if needed.

andrewgodwin:
Yeah, I don't mind vendoring (copying) here - it's small enough and the license is directly compatible that we're fine, and that regex doesn't exactly change every month!

andrewgodwin:
Thanks for this!

PR #117, opened 2022-12-05
Only cache unauthenticated page views

manfre:


manfre:
It does not. Fixing

core/decorators.py
```diff
@@ -1,34 +1,81 @@
...
+    if (
+        "application/json" in accept
+        or "application/ld" in accept
+        or "application/activity" in accept
+    ):
+        return "accept-ap"
+    elif "text/html" in accept:
+        return "accept-html"
+    # Fallback is potentially uncached
+    return slugify("".join(str(x) for x in request.accepted_types))
```
manfre, line 29: @andrewgodwin I'm not sure how I feel about this fallback. We need to cache before a decision is made and the `Accept` header can be very arbitrary. Good enough or any suggestions on how to improve this?

andrewgodwin: Thinking about this, we should just have a middleware that detects the accept header or not. Let me write one of those and it'll simplify things, as we only care about "AP or not", versus the true power of the Accept header.

users/views/identity.py
```diff
@@ -10,15 +10,17 @@
...
 from core.ld import canonicalise
 from core.models import Config
 from users.decorators import identity_required
 from users.models import Domain, Follow, FollowStates, Identity, IdentityStates
 from users.shortcuts import by_handle_or_404
 
 
-@method_decorator(per_identity_cache_page(), name="dispatch")
+@method_decorator(
+    cache_page_by_ap_json(public_only=True, key_prefix="asdf"), name="dispatch"
```
andrewgodwin, line 22: Is this asdf intentional?

manfre: No, removed it.

andrewgodwin (approved):

PR #123, opened 2022-12-06
Collapse linkify mentions

manfre:
Makes linkify shortening a bit more aware of short name collisions.


andrewgodwin (approved):
Was momentarily unsure about only making one name longer, but yeah, let's do it.

PR #130, opened 2022-12-06
Improve the job names of GitHub Actions

shuuji3:
When I first see the Actions page, I'm a bit confused as there are the same two py3.10/py3.11 jobs. 

This tweaks the name of each `test` job by adding `with postgresql` or `with sqlite` string at the end.

**before**
Screenshot 2022-12-06 at 17 32 19

**after**
Screenshot 2022-12-06 at 17 31 58

You can see the test run here: https://github.com/shuuji3/takahe/actions/runs/3627857982

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   test:
-    name: test py${{ matrix.python-version }}
+    name: test py${{ matrix.python-version }} with ${{ matrix.db_name }}
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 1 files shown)


shuuji3:
Ah, you're right. I misunderstood that keys under `include` cannot be used in the `name` value but that was not the case.

I amended to keep the original `db` values and simply inserted `db_name`. Is this what you meant? @TkTech

andrewgodwin:
This looks good enough to merge to me - if there's any further tweaks, we can always do them in a new PR!

PR #131, opened 2022-12-06
Don't cache pip packages when building docker image

dcwatson:
Disabling the pip cache significantly slims down the docker image (by about 70mb in my local testing). I think the pip cache is likely not even used, since the layer that does the install will either be cached or re-built entirely.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -19,7 +19,7 @@ RUN apt-get update \
         # Required to build lxml on arm64.
         libxslt1-dev \
         zlib1g-dev \
-    && python3 -m pip install --upgrade -r requirements.txt \
+    && python3 -m pip install --no-cache-dir --upgrade -r requirements.txt \
     && apt-get purge -y --auto-remove \
         gcc \
         libc6-dev \


abkfenris:
Using buildkit it's possible to both exclude caches from the final image, while also using them if available to speed up rebuilding locally. [Here is some info on using it in Python](https://pythonspeed.com/articles/docker-cache-pip-downloads/)

I asked pip for the cache directory in the image

```bash
root@7bea115738ac:/takahe# python -m pip cache dir
/root/.cache/pip
```

So I think the change would be adding a cache mount to the beginning of the RUN command, plus placing `#syntax=docker/dockerfile:1.2` at the top of the file.

```dockerfile
RUN --mount=type=cache,id=pip,target=/root/.cache apt-get update \
    && apt-get install -y --no-install-recommends \
        gcc \
        libc6-dev \
        libpq-dev \
        # Required to build lxml on arm64.
        libxslt1-dev \
        zlib1g-dev \
    && python3 -m pip install --upgrade -r requirements.txt \
    && apt-get purge -y --auto-remove \
        gcc \
        libc6-dev \
        libpq-dev \
        libxslt1-dev \
        zlib1g-dev \
    && rm -rf /var/lib/apt/lists/*
```

It should also be possible to cache apt.

PR #134, opened 2022-12-06
'View in admin' link for post popup menu

manfre:
* Fixes styling on menu to keep `View Post & Replies` on a single line
* Adjust `z-index` of menu so it shows over post links of the next post
* Admins will see a "View In Admin" link for posts to quickly navigate to the admin for troubleshooting/intervention

image


andrewgodwin:
I'm not sure how much we want to direct admins to the Django Admin - we can swap this to a moderation view when things are further along, maybe.

Issue #136, opened 2022-12-06
[Accessibility iOS VoiceOver]


whatSocks:
I tried navigating https://takahe.social/ a goal of finding the local feed using only iOS VoiceOver on an iPhone.

Main takeaways:

- It wasn't obvious to me that I have to swipe right-left twice to get to the navigation
- The banner image has a nice filename so that was obviously a banner image
- the images under "People" don't have as nice names, it would be nice to have a fall back like <username's profile image>
- using "rotor" to search of lists or tables, it didn't find the People section, which visually seems to be a list with 2 items
- the links on the sidebar could be more buttony 

It's much easier to navigate and read posts on Metatext (what i use for Mastodon) but it's hard for me to enunciate what makes that app more usable.

PR #137, opened 2022-12-06
Add an alt label identity images.

TkTech:
Resolves 3rd bullet point in #136.


manfre:
https://www.w3.org/WAI/tutorials/images/

`alt=""` might be more appropriate.

PR #145, opened 2022-12-07
Set link to user icon in identity panel

### LINKED ISSUE #144: Cannot click the profile icon in the identity panel
Currently, we cannot click the profile icon in the identity panel. This is different from the behavior on the timeline profile icon, causing failure to navigate to the expected page, especially in a small window.

Screenshot 2022-12-07 at 23 28 29

shuuji3:
Resolves #144


andrewgodwin:
Thanks

PR #152, opened 2022-12-10
Add the character counter next to the Post button

### LINKED ISSUE #14: 500 on add profile pic to identity via frontend

### LINKED ISSUE #146: Feature request: character count under/near “compose” textarea
I find it inconvenient to not “know” as I’m composing how close I am to the configured post character limit.

My Mac has some tools set up that I could use (“services”) to check the character count, but sometimes I forget, and it’s a bit awkward. Same on my phone, I have a way to check the character count before submitting the form, but it’s cumbersome.

---

A few related ancillary feature requests:

- [ ] if a draft post exceeds the configured character limit, use client-side HTML form validation (`maxlength`) to prevent the form from being submitted when it’s invalid (faster feedback loop, less server load, works offline, etc, etc)
- [ ] if the compose form includes an image attachment, and it’s submitted, and is found to be invalid because the post has too many characters (or for any other reason) let’s not lose that image attachment; let’s retain it (this could be reframed as a bug, I suppose)

(I’ve made the above list a checklist because I suspect that might make it easier to break them out into discrete issues, if so desired.)

---

These are all things I personally could probably contribute, but I don’t have time at this moment, so I wanted to at least just jot these thoughts down and share them.

shuuji3:
resolves #146

I couldn't find a good place to put the counter near the <textarea> so put it next to the Post button.

Here's how this counter works:

https://user-images.githubusercontent.com/1425259/206854741-b02599f3-1d1a-453c-b452-c25fbefa6a1c.mov

### DIFF
--- a/activities/views/compose.py
+++ b/activities/views/compose.py
@@ -28,6 +28,19 @@ class Compose(FormView):
             widget=forms.Textarea(
                 attrs={
                     "placeholder": "What's on your mind?",
+                    "_": f"""
+                    on load or input
+                        set characters to my.value.trim().length
+                        put { Config.system.post_length } - characters into #character-counter
+
+                        if characters > { Config.system.post_length } then
+                            set #character-counter's style.color to 'var(--color-text-error)'
+                            add [@disabled=] to #post-button
+                        else
+                            set #character-counter's style.color to ''
+                            remove @disabled from #post-button
+                        end
+                    """,
                 },
             )
         )
--- a/activities/views/timelines.py
+++ b/activities/views/timelines.py
@@ -20,6 +20,19 @@ class Home(FormView):
             widget=forms.Textarea(
                 attrs={
                     "placeholder": "What's on your mind?",
+                    "_": f"""
+                    on load or input
+                        set characters to my.value.trim().length
+                        put { Config.system.post_length } - characters into #character-counter
+
+                        if characters > { Config.system.post_length } then
+                            set #character-counter's style.color to 'var(--color-text-error)'
+                            add [@disabled=] to #post-button
+                        else
+                            set #character-counter's style.color to ''
+                            remove @disabled from #post-button
+                        end
+                    """,
                 },
             )
         )
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -93,11 +93,13 @@ td a {
     --color-text-dull: #99a;
     --color-text-main: #fff;
     --color-text-link: rgb(176, 194, 206);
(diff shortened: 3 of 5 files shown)


shuuji3:
I couldn't fix the tests. I would appreciate it if someone could help me.

activities/views/compose.py
```diff
@@ -28,6 +28,19 @@ class form_class(forms.Form):
             widget=forms.Textarea(
                 attrs={
                     "placeholder": "What's on your mind?",
+                    "_": f"""
+                    on load or input
+                        set characters to my.value.trim().length
+                        put { Config.system.post_length } - characters into #character-counter
```
shuuji3, line 34: It looks working but I'm not confident this is the correct way to get `post_length`.

andrewgodwin: So this is the right place to pull it from, but the wrong _time_ to pull it; you're doing this in a place that is run when the module is loaded, but it should be done each request instead (this is why the tests are failing).

You will, instead, have to either add this attribute when the form is created in its __init__, or move it into the template with a special override for this specific textarea.

I'm happy to do this for you if you'd like.

shuuji3: Thanks for taking a look and giving hints🙂 Hmm, it seems like I need to understand the lifecycle of these components better. I'll try to find a right way again tomorrow.

shuuji3: @andrewgodwin Ah sorry, I misunderstood your last comment when I read it at midnight. As it still seems complicated for me to amend the change, could you fix the error if you can? Thanks.

andrewgodwin: Sure thing - I'll just merge this in as-is and then immediately fix it, rather than pushing changes to your branch (I'm feeling lazy this morning).

Thanks for figuring out the hyperscript!

shuuji3: No problem at all and thanks! Yeah, takahē is how I knew the hyperscript at first but writing hyperscript is a kind of fun. It reminds me to write AppleScript :)

andrewgodwin:
You can see the fix I did here, moving it into `__init__`: https://github.com/jointakahe/takahe/commit/f504e9c3516db48647b75477f4672e288413daf7

I also took the opportunity to make both places use the same form, since you reminded me they were needlessly copied!

shuuji3:
I see, now I understood how I should have amended the field attribute. Yes, I noticed the duplications, nice refactoring!

Issue #58, opened 2022-11-26
Following/Followers counts on local profile pages


andrewgodwin:
Along with the option to hide them in profile settings

PR #159, opened 2022-12-12
Show follows and following counts on profile page

vibragiel:
Fixes https://github.com/jointakahe/takahe/issues/58

Hi, this is the first time in 7 years that I'm doing some Django, so bear with me :)

This PR adds a small section in the profile page to show the follows and following counts of the identity.

image

It also lets the user configure the visibility of this.

image

This is missing:
* A cache for these stats, so two extra database hits are done per profile visit
* A count for total toots/posts, which would be nice


andrewgodwin:
Looks like a good start!

For the "visible follows or not" option, that shouldn't be a new field on Identity, but instead an entry in our identity Config system. Look for "toot_mode" as a setting that you can see where it's defined and how its setting is added to the page.

vibragiel:
Sounds more reasonable! Let me fix that.

vibragiel:
@andrewgodwin I created a new "Privacy" page, because it doesn't belong to the Profile but it's also not an Interface thing, since it's not actually configuring how the user experiences the interface, but what others are able to see.

What do you think?

image

manfre:
@vibragiel This seems like it should go on the `Profile` page in the `Details` section (below `Discoverable` makes sense to me). This feature controls the content that appears with the `Identity` profile.

andrewgodwin:
Yes, this belongs next to Discoverable. You'll need to do a little custom coding on that page rather than being able to use a generic Settings page, but that really is where it belongs.

vibragiel:
Hmm. My reasoning for not putting it there is that this feels a bit different from the rest of the profile stuff, doesn't it? It's not configuring "who you are" but rather it depends on your past activity. Besides, it so happens that initially these counts appear with the profile info (or below it), but information about your follows could easily creep into other parts of the UI or the API. That's why I put it in a Privacy section, because this is "I don't want my follows to be public, wherever they appear".

vibragiel:
Fair enough! I moved it into the Profile page.
image

andrewgodwin (approved):

andrewgodwin:
Lovely, thanks for your work on this!

TkTech:
Should I just change the wording of this setting to also apply to showing the list of followers, or add a new setting?

"Whether or not to show your following and followers in your profile"

andrewgodwin:
Let's change the wording and hide the follower lists behind it too. We should probably also have the client API return 0 or None for these numbers too in that case (not sure what Mastodon does, their client API schema is not fully correct)

PR #160, opened 2022-12-14
FormOrJsonParser api.views.oauth

manfre:
Caught this due to https://pinafore.social using `application/json` flow.


andrewgodwin (approved):
Ahhh yes, should have thought about this happening here too.

PR #162, opened 2022-12-14
Validate token scope and return 401 for invalid clients

manfre:
Pinafore.social got stuck in a broken state that I was able to clear by return 401s, instead of 400s when the client_id was invalid. This matches the mastodon docs https://docs.joinmastodon.org/methods/oauth/#401-unauthorized

Also added token scope comparison.

### DIFF
--- a/api/views/oauth.py
+++ b/api/views/oauth.py
@@ -59,15 +59,6 @@ class AuthorizationView(LoginRequiredMixin, TemplateView):
                 status=401,
             )
 
-        if not application.is_scope_subset(scope):
-            return JsonResponse(
-                {
-                    "error": "invalid_scope",
-                    "error_description": "The requested scope is invalid, unknown, or malformed.",
-                },
-                status=400,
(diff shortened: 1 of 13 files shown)


api/views/oauth.py
```diff
@@ -47,10 +47,32 @@ def post(self, request):
...
+        except Application.DoesNotExist:
+            return JsonResponse(
+                {
+                    "error": "invalid_client",
+                    "error_description": "Client authentication failed due to unknown client, no client authentication included, or unsupported authentication method.",
+                },
+                status=401,
+            )
+
+        if not application.is_scope_subset(scope):
```
andrewgodwin, line 62: So when I used Tusky, it didn't set the application scope at all and just relied on the authentication scope, from what I can see. I don't see any security advantage from respecting the application scope here - let's just go with whatever they request.

The scope subset code should probably move to Token so we can use it for view auth checking.

Issue #161, opened 2022-12-14
Simple photo viewer


shuuji3:
The current implementation converts images to WebP format. While WebP itself is great, some browsers such as Google Chrome and Safari (but not Firefox) always download it every time clicking the image, instead of opening it in the browser. It's inconvenient when we want to see enlarg images.

To prevent this a bit annoying download behavior, I'd like a simple photo viewer. I'm thinking about something like this: https://pixelfed.social/p/PixFed/89129948895383552. A similar one can be implemented in _hyperscript.

andrewgodwin:
Yes, that was the intention - just a basic popover box with an X will do.

TkTech:
> The current implementation converts images to WebP format. While WebP itself is great, some browsers such as Google Chrome and Safari (but not Firefox) always download it every time clicking the image, instead of opening it in the browser. It's inconvenient when we want to see enlarg images.

This is resolved for NEW uploads on instances using S3, as we now properly set the `Content-Type`  and `Content-Disposition` headers, which were previously defaulting to octet. Local storage and GCS-backed instances are still borked.

PR #163, opened 2022-12-14
Set the content type and disposition of webp files uploaded to S3 to …

TkTech:
…display them inline in chrome instead of downloading them. See #161.

Also, small bump to the patch version number of the Python used for heroku due to a security fix.


shuuji3:
Thank you for the quick fix! 👍🏻 I created a issue #164 for Google Cloud Storage.

PR #170, opened 2022-12-15
Several pagination improvements

### LINKED ISSUE #126: Add pagination to Home timeline

### LINKED ISSUE #168: Paginate Follows page
Also probably wants a search function, and filters

vibragiel:
This PR adds pagination to the Home timeline and the Follows page.

A couple of caveats:

* The Follows were ordered by the username, but we can't do that efficiently now without traversing the whole unpaginated queryset (which would defeat the purpose of pagination). They're now ordered by `id`, which _should_ equate to creation order.
* I removed a lot of slices in the paginated querysets. I don't know if those were deliberate or leftovers from before pagination was added.
* I added buttons to go back to previous pages. Again, I don't know if that omission was deliberate.

Fixes  #126 and partially fixes #168.

### DIFF
--- a/activities/views/follows.py
+++ b/activities/views/follows.py
@@ -1,31 +1,33 @@
-from django.core.paginator import Paginator
 from django.db.models import Q
 from django.utils.decorators import method_decorator
-from django.views.generic import TemplateView
+from django.views.generic import ListView
 
 from users.decorators import identity_required
 from users.models import Follow, FollowStates
 
 
 @method_decorator(identity_required, name="dispatch")
-class FollowsPage(TemplateView):
+class Follows(ListView):
     """
     Shows followers/follows.
     """
 
     template_name = "activities/follows.html"
+    extra_context = {
+        "section": "follows",
+    }
+    paginate_by = 50
 
-    def get_context_data(self):
-        # Gather all identities with a following relationship with us
-        follows = Follow.objects.filter(
+    def get_queryset(self):
+        return Follow.objects.filter(
             Q(source=self.request.identity) | Q(target=self.request.identity),
             state__in=FollowStates.group_active(),
-        ).order_by("-id")
-        paginator = Paginator(follows, 50)
-        page_number = self.request.GET.get("page")
-        page_obj = paginator.get_page(page_number)
+        ).order_by("-created")
+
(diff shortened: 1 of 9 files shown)


activities/views/follows.py
```diff
@@ -15,19 +17,25 @@ class FollowsPage(TemplateView):
...
-            state__in=FollowStates.group_active()
-        ):
-            identities.setdefault(outbound_follow.target, {})[
-                "outbound"
-            ] = outbound_follow
-        for inbound_follow in self.request.identity.inbound_follows.filter(
-            state__in=FollowStates.group_active()
-        ):
-            identities.setdefault(inbound_follow.source, {})["inbound"] = inbound_follow
+        follows = Follow.objects.filter(
```
andrewgodwin, line 20: Could you move this to a ListView instead? See the "federation" admin page for a good example.

activities/views/follows.py
```diff
@@ -15,19 +17,25 @@ class FollowsPage(TemplateView):
...
+            Q(source=self.request.identity) | Q(target=self.request.identity),
+            state__in=FollowStates.group_active(),
+        ).order_by("-id")
```
andrewgodwin, line 23: There's a "created" on Follow too, let's order by that instead.

activities/views/follows.py
vibragiel: Done at https://github.com/jointakahe/takahe/pull/170/commits/265ff7c50f3c261ef858fa61c46a0466dea5d49c

activities/views/follows.py
vibragiel: Done at https://github.com/jointakahe/takahe/pull/170/commits/265ff7c50f3c261ef858fa61c46a0466dea5d49c

vibragiel:
Take a look, @andrewgodwin. The only thing is that follows are now treated individually, so if user A follows B, and B follows A, that will show as two separate items in the list. I don't see an obvious and efficient way of fixing that without some kludgy SQL.

andrewgodwin (approved):

andrewgodwin:
No problem - I think I know a way to maybe fix that, I'll merge it in and see what I can accomplish. Thanks for doing this!

PR #171, opened 2022-12-15
Set the correct `Content-Type` header for static WebP images

shuuji3:
This is a simiar fix as #163 but for the local WebP image serving. Previously, `content-deposition` has already been set to `inline` by Django, but `content-type` was `application/octet-stream`, which was causing downloading.

It can be implemented by customizing the original `django.views.static.serve` view for `*.webp` files.

## Reason for custom view

The reason why we cannot use the original `serve` is that [the current Django implementation](https://cs.github.com/django/django/blob/09ffc5c1212d4ced58b708cbbf3dfbfb77b782ca/django/http/response.py#L558) set the `Content-Type` header based on the return value of `mimetypes.guess_type()`. Although the `image/webp` MIME type has been added to `mimetypes` database just since Python 3.11 (ref. python/cpython#29259), the `image/webp` is not yet registered to [the official IANA mime-type list](https://www.iana.org/assignments/media-types/media-types.xhtml) and `guess_type()` in the Django source code needs to have `strict=False` option to guess `image/webp`, but it doesn't.

## Screenshot after this change
Screenshot 2022-12-16 at 1 18 26

Screenshot 2022-12-16 at 1 17 57

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -4,6 +4,7 @@ from django.templatetags.static import static
 from django.utils.decorators import method_decorator
 from django.utils.safestring import mark_safe
 from django.views.generic import TemplateView, View
+from django.views.static import serve
 
 from activities.views.timelines import Home
 from core.decorators import cache_page
@@ -89,3 +90,14 @@ class FlatPage(TemplateView):
             "title": self.title,
             "content": mark_safe(html),
         }
+
+
+def custom_static_serve(*args, **keywords):
+    """
+    Set the correct `Content-Type` header for static WebP images
+    since Django cannot guess the MIME type of WebP images.
+    """
+    response = serve(*args, **keywords)
+    if keywords["path"].endswith(".webp"):
+        response.headers["Content-Type"] = "image/webp"
+    return response
--- a/takahe/urls.py
+++ b/takahe/urls.py
@@ -1,7 +1,6 @@
 from django.conf import settings as djsettings
 from django.contrib import admin as djadmin
 from django.urls import path, re_path
-from django.views.static import serve
 
 from activities.views import compose, explore, follows, posts, search, timelines
 from api.views import api_router, oauth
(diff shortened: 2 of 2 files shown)


shuuji3:
The error messages in failed tests seem to be the same ones by the last commit to the main branch.

At least I confirmed to pass all tests in the local container:

```shell
root@f23f04b91240:/takahe# pytest
========================================================================= test session starts ==========================================================================
platform linux -- Python 3.11.0, pytest-7.2.0, pluggy-1.0.0
django: settings: takahe.settings (from option)
rootdir: /takahe, configfile: setup.cfg
plugins: dotenv-0.5.2, django-4.5.2, httpx-0.21.2, asyncio-0.20.3, anyio-3.6.2
asyncio: mode=Mode.STRICT
collected 54 items                                                                                                                                                     

tests/activities/models/test_post.py ....                                                                                                                        [  7%]
tests/activities/models/test_post_targets.py ...                                                                                                                 [ 12%]
tests/activities/views/test_compose.py ..                                                                                                                        [ 16%]
tests/activities/views/test_posts.py .                                                                                                                           [ 18%]
tests/activities/views/test_timelines.py .                                                                                                                       [ 20%]
tests/users/models/test_identity.py ...                                                                                                                          [ 25%]
tests/users/views/test_activitypub.py ...                                                                                                                        [ 31%]
tests/users/views/test_auth.py ....                                                                                                                              [ 38%]
stator/tests/test_graph.py ...                                                                                                                                   [ 44%]
tests/activities/models/test_hashtag.py ..                                                                                                                       [ 48%]
tests/activities/templatetags/test_activity_tags.py ..                                                                                                           [ 51%]
tests/core/test_html.py ..                                                                                                                                       [ 55%]
tests/core/test_ld.py .                                                                                                                                          [ 57%]
tests/core/test_signatures.py ....                                                                                                                               [ 64%]
tests/users/views/test_domains.py ...................                                                                                                            [100%]

========================================================================== 54 passed in 9.72s ==========================================================================
r
```

andrewgodwin (approved):

andrewgodwin:
Thanks for figuring this out!

PR #182, opened 2022-12-17
Post types

manfre:
Additional post types are handled using pydantic models to store the arbitrary data for each type. Specialized types will render their contents using templates, if they're found. If not, it will display a generic "See original" link.


andrewgodwin (approved):

andrewgodwin:
Looks good - I'll remove the "question" type drop as well and let it feed into this.

PR #184, opened 2022-12-17
icon and image for Identity could be a list

manfre:
Pixelfed (and others) allow more than one picture for Identities.


users/models/identity.py
```diff
@@ -424,6 +424,19 @@ async def fetch_webfinger(cls, handle: str) -> tuple[str | None, str | None]:
                 return link["href"], data["subject"]
         return None, None
 
+    def _first_image_url(self, data) -> str | None:
```
andrewgodwin, line 427: Can you move this into `core.ld`, like the other JSON-LD utility function are?

andrewgodwin:
Thanks!

PR #193, opened 2022-12-18
Increase the allowed size of Post.object_uri and Post.url from 500 ch…

TkTech:
…aracters to a more reasonable 2048. See https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers.

### DIFF
--- a/activities/models/post.py
+++ b/activities/models/post.py
@@ -180,7 +180,12 @@ class Post(StatorModel):
     local = models.BooleanField()
 
     # The canonical object ID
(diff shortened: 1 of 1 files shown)


andrewgodwin:
I believe this technically needs a migration, even though PostgreSQL won't actually change anything?

andrewgodwin:
Actually nevermind, I'll bundle the migration into the one I want to do for TimelineEvent.published

PR #196, opened 2022-12-18
Identity fanout

manfre:
Identity update/delete fanout.


users/admin.py
```diff
@@ -75,8 +75,8 @@ class FollowAdmin(admin.ModelAdmin):
     list_filter = [LocalSourceFilter, LocalTargetFilter, "state"]
     raw_id_fields = ["source", "target"]
 
-    def has_add_permission(self, request, obj=None):
-        return False
+    # def has_add_permission(self, request, obj=None):
+    #     return False
```
manfre, line 79: This was done for ad-hoc testing. Will clean up.

users/models/identity.py
```diff
@@ -32,19 +32,71 @@
...
+
+    When a local identity is "edited" or "deleted", it will fanout the change to
+    all followers and transition to "updated"
     """
 
     outdated = State(try_interval=3600, force_initial=True)
     updated = State(try_interval=86400 * 7, attempt_immediately=False)
 
+    edited = State(try_interval=300, attempt_immediately=True)
+    # deleted = State(externally_progressed=True)
```
manfre, line 46: Not sure if we want a separate delete or let it flow through as an edit that reacts to `Identity.deleted`

andrewgodwin: Separate delete, as that way we can also use the state as a "is this user deleted" filter

andrewgodwin (commented):
Looks good apart from the two areas you highlighted

users/models/identity.py
```diff
@@ -30,19 +30,71 @@
...
+
+            await FanOut.objects.acreate(
+                identity=follower.source,
+                type=type_,
+                subject_identity=identity,
+            )
+            shared_inboxes.add(shared_uri)
+
+    @classmethod
+    async def handle_new(cls, instance):
```
andrewgodwin, line 76: There's no "new" state here, so why is this handler here?

users/models/identity.py
```diff
@@ -30,19 +30,71 @@
...
+    @classmethod
+    async def handle_edited(cls, instance: "Identity"):
+        from activities.models import FanOut
+
+        if not instance.local:
+            return cls.updated
+
+        identity = await instance.afetch_full()
+
+        if identity.deleted:
```
andrewgodwin, line 88: While I did leave a "deleted" field on the model, I'd rather this follows the existing Post pattern where we actually use "deleted" and "deleted_fanned_out" states, and we ignore the "deleted" field on the model entirely (and remove it when the next migration opportunity arises)

andrewgodwin (commented):
Couple of small requests. Also worth noting - Mastodon fans out profile deletes to every single server it knows about (presumably in case they are cached locally), but I don't know if we need to do the same yet.

users/models/identity.py
manfre: Remnants after a rebase.

users/models/identity.py
manfre: Changed flow to `deleted` and `deleted_fanned_out`.

andrewgodwin (approved):

PR #207, opened 2022-12-19
Mitra Fixes

TkTech:
Mitra has a bug in its handling of the `Accept` header (https://f.tkte.ch/@tktech@tkte.ch/posts/13530/) and this change doesn't seem to affect any of the existing servers.  This fixes search-by-url breaking on Mitra because it would return HTML on /post/ URLs instead of redirecting them to /objects/.

Re-allows `<a>` in posts, which was (accidentally?) removed in https://github.com/jointakahe/takahe/commit/2a3690d1c148da5dd799052403ba7290e1fb7de0.

### DIFF
--- a/core/html.py
+++ b/core/html.py
@@ -20,7 +20,7 @@ def sanitize_post(post_html: str) -> str:
     Only allows a, br, p and span tags, and class attributes.
     """
     cleaner = bleach.Cleaner(
-        tags=["br", "p"],
+        tags=["br", "p", "a"],
         attributes={  # type:ignore
             "a": allow_a,
             "p": ["class"],
--- a/core/signatures.py
+++ b/core/signatures.py
@@ -200,7 +200,7 @@ class HttpSignature:
             body_bytes = b""
         # GET requests get implicit accept headers added
         if method == "get":
-            headers["Accept"] = "application/activity+json, application/ld+json"
+            headers["Accept"] = "application/activity+json"
         # Sign the headers
         signed_string = "\n".join(
             f"{name.lower()}: {value}" for name, value in headers.items()
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,6 +3,8 @@ ARG IMAGE_LABEL=3.11.1-slim-buster
 
 FROM ${IMAGE_HOST}:${IMAGE_LABEL}
 
+ENV PYTHONUNBUFFERED=1
+
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         libpq5 \


andrewgodwin (approved):
I was wondering when someone would choke on the accept header!

silverpill:
1. Your `Accept` header is not correct. Here's what ActivityPub spec says about retrieving objects:

>The client MUST specify an Accept header with the `application/ld+json; profile="https://www.w3.org/ns/activitystreams"` media type in order to retrieve the activity. 

-- https://www.w3.org/TR/activitypub/#retrieving-objects

2. `/post/` paths are not object IDs and they belong to a static frontend. You shouldn't expect these URLs to redirect to AP representations (although I agree that this is a nice trick). But the backend should handle object retrieval requests even if `Accept` header contains some unusual value such as  `application/activity+json, application/ld+json`.

3. The actual reason why Takahe isn't federating with Mitra is the lack of `outbox` in your actor objects. This is a violation of the AcitivityPub spec:

>Actor objects MUST have, in addition to the properties mandated by 3.1 Object Identifiers, the following properties:
>outbox
>An ActivityStreams OrderedCollection comprised of all the messages produced by the actor; see 5.1 Outbox.

-- https://www.w3.org/TR/activitypub/#actor-objects

Outbox URL can simply return empty array, but the `outbox` property must be present in actor object.

andrewgodwin:
I have been very much going with "well what do servers _actually_ do" rather than what the spec says, but I appreciate the pointers. It's funny that nothing else has needed outbox until now, but it'll be easy to add in.

andrewgodwin:
Also, it's worth noting that ActivityStreams says "Because Activity Streams 2.0 can be considered a restricted profile of JSON-LD, Implementations should consider the `application/ld+json; profile="https://www.w3.org/ns/activitystreams"` media type as being equivalent to `application/activity+json`."), so I think both need to be supported.

andrewgodwin:
There, takahe.social now has an Outbox for all local actors and it even has Note objects in it, plus the accept header is now being sent exactly as in the spec. Hopefully this will allow Mitra federation.

silverpill:
Yes, everything seems to be working now.

>the accept header is now being sent exactly as in the spec

Several Fediverse platforms still don't return object representation when `Accept` header contains `application/ld+json; profile="https://www.w3.org/ns/activitystreams"`, but these projects are mostly abandonware.
I don't care about them so I use `application/ld+json; profile="https://www.w3.org/ns/activitystreams"`. It's better for the ecosystem: one less quirk to learn about for new developers (who often complain that their software doesn't federate when they implement everything according to a spec).

andrewgodwin:
Yes, I only care about currently active servers, so unless one of those is unhappy I think this is a good place to be. We'll still accept any variation inbound for now, though, since that's easier to be permissive on.

PR #213, opened 2022-12-20
Fix issues with hashtags being linkified when at the end of a URL (li…

TkTech:
…ke https://github.com/TkTech/notifico/blob/master/README.md?plain=1#L12) and other places they shouldn't be.

### DIFF
--- a/activities/models/hashtag.py
+++ b/activities/models/hashtag.py
@@ -14,7 +14,7 @@ from core.html import strip_html, hashtag_callback, ALLOWED_HTML_TAGS
 from core.models import Config
 from stator.models import State, StateField, StateGraph, StatorModel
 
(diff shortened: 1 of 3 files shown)


andrewgodwin (approved):

TkTech:
This is not mergable as-is, because it cannot support hashtags split over multiple HTML tags which apparently Mastodon does.

andrewgodwin:
OK, I will see if I can think of a different way later today.

TkTech:
Mastodon appears to be using several different regexes in multiple stages, but I hesitate to look to closely as it's GPL.

andrewgodwin:
I don't believe Mastodon supports preserving <a> hrefs directly, though, which helps a lot in this regard. I'm going to revert us to the same situation (only linkifying direct URLs, mentions, and hashtags) and we can figure out how to safely preserve actual links from other servers later on.

PR #230, opened 2022-12-22
Various UX tweaks and front end fixes.

### LINKED ISSUE #225: Long URLs escape the post box in UI
We need a way to wrap them or truncate them.

TkTech:
- Add a nginx.dev.conf which is used by the dockerfile when django is in debug mode. This fixes issues with static assets not being available unless collectstatic is run even in debug mode.

Fixes #225


TkTech:
Tweaks the behaviour of links in post boxes to match the hashtags (no underline, underline on hover) and switches their color to the Takahe highlight color to better standout as there were comments on the contrast being too low.

image

TkTech:
Ultra-compact view on small screens:
image

TkTech:
Moved action buttons to the bottom, which gives us a lot more room to work with on screens small and large, and is more typical of what you'd see on Twitter or Mastodon.

image

TkTech:
Posts now show the total # of boosts and likes next to the action button.

Caveat, clicking the button doesn't immediately refresh the count. I don't really want to put the effort in until #231 is closed.

image

@andrewgodwin this is ready for review.

whatSocks:
@TkTech I can't tell from the screenshots but when I'm testing on mobile (so an actual small screen) I sometimes want to see what version I'm on but it's hard to for me to see if the page I'm on even shows this information. Could this be displayed in the right side menu or someplace more accessible?

docker/nginx.conf
```diff
@@ -9,7 +9,7 @@ http {
 
     include /etc/nginx/mime.types;
 
-    proxy_cache_path /cache/nginx levels=1:2 keys_zone=takahe:20m inactive=14d max_size=__CACHESIZE__;
+    proxy_cache_path /cache/nginx levels=1:2 keys_zone=takahe:20m inactive=14d max_size=1g;
```
andrewgodwin, line 12: Why did you remove the dynamic cache size replacement? This should be configurable.

docker/nginx.dev.conf
```diff
@@ -0,0 +1,98 @@
+daemon off;
```
andrewgodwin, line 1: Is the purpose of this entire file to remove /static/? If so, did you try `try_files` first, where the main static alias tries the local file first and then goes to proxy_pass?

andrewgodwin (commented):
I have some questions about the nginx changes :)

As for the actions being at the bottom - I was trying to avoid using that vertical space, but I guess we don't have much option at some point. Lots more vertical space than horizontal.

docker/nginx.dev.conf
TkTech: I did not, not super familiar with nginx. And yes, it's to remove /static/

andrewgodwin (approved):

PR #242, opened 2022-12-23
Better looking image galleries, video attachment player.

TkTech:
Posts will multiple images look like this:

image


TkTech:
Closing in favor of #244.

PR #245, opened 2022-12-23
Handle more also_known_as scenarios

manfre:
Acquired Domains:

* Handle "/user/{username}" and "/users/{username}" for any Domain.local
* Domain.local must exist and an Identity must have also_known_as.
* Identity.domain doesn't need to match the matched local Domain

Incorrect actor_uri:

Commit #578b504c924176e66531cd0ce64120de65e80c7a added the short form actor_uri as a webfinger alias, but InboxMessage only supports lookup by actor_uri. This adds also_known_as lookups in fetch_actor with Sentry messaging when conflicts are found.

### DIFF
--- a/users/models/identity.py
+++ b/users/models/identity.py
@@ -10,7 +10,7 @@ from django.template.defaultfilters import linebreaks_filter
 from django.utils import timezone
 from django.utils.functional import lazy
 
-from core.exceptions import ActorMismatchError
+from core.exceptions import ActorMismatchError, capture_message
 from core.html import ContentRenderer, strip_html
 from core.ld import (
     canonicalise,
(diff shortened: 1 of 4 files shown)


users/models/identity.py
```diff
@@ -256,8 +256,8 @@ def all_absolute_profile_uris(self) -> list[str]:
         if not self.local:
             return [self.profile_uri]
         return [
-            f"https://{self.domain.uri_domain}/@{self.username}/",
             f"https://{self.domain.uri_domain}/@{self.username}@{self.domain_id}/",
+            f"https://{self.domain.uri_domain}/@{self.username}/",
```
manfre, line 260: The ordering here is likely the source of some of the actor_uri issues for Takahe servers.

PR #247, opened 2022-12-24
Fixes #239 - Add option to hide Boost and Like counts

### LINKED ISSUE #239: [Feature Wish] Hide Like/Repost counter
Maybe identity-based hiding of like and repost counter, because I think, counts are not really healthy.

manfre:


core/context.py
```diff
@@ -2,10 +2,15 @@
 
 
 def config_context(request):
+    config_identity = request.identity.config_identity if request.identity else None
+
+    reaction_number_css = ""
+    if config_identity and not config_identity.visible_reaction_counts:
+        reaction_number_css = "display: none;"
```
andrewgodwin, line 9: This shouldn't be done with a CSS injection like this - instead, there should be a dynamic `<style>` block in the head of base.html that does this instead based on classes.

manfre: I like that approach a lot better!

andrewgodwin (approved):

PR #250, opened 2022-12-24
Identity background images can be relative

manfre:
Fixes the surprise server profile bomb for those using local media storage.

### DIFF
--- a/users/models/identity.py
+++ b/users/models/identity.py
@@ -276,7 +276,7 @@ class Identity(StatorModel):
         Returns a background image for us, returning None if there isn't one
         """
         if self.image:
-            return RelativeAbsoluteUrl(self.image.url)
+            return AutoAbsoluteUrl(self.image.url)
         elif self.image_uri:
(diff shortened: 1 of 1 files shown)


users/models/identity.py
```diff
@@ -276,7 +276,7 @@ def local_image_url(self) -> RelativeAbsoluteUrl | None:
         Returns a background image for us, returning None if there isn't one
         """
         if self.image:
-            return RelativeAbsoluteUrl(self.image.url)
+            return AutoAbsoluteUrl(self.image.url)
```
andrewgodwin, line 279: I think this will break if the URI is already absolute?

manfre: My changes makes `local_image_url` match the behavior of `local_icon_url`. icon is more likely to break sites and no one has complained, so I believe that is the correct way of doing it. Fewer people upload background images at this point.

andrewgodwin: Yeah, on closer inspection the urljoin should respect an absolute second half, so this must work.

PR #255, opened 2022-12-24
Save changes made by pre-commit in the test workflow.

TkTech:
Also enables the pre-commit cache as suggested by pre-commit.


TkTech:
It works, but I don't really like _how_ it works. Canning for now.

PR #265, opened 2022-12-25
Fixed mention linking with mixed case usernames

manfre:
This preserves whatever case they used and continues to do the username conflict detection case insensitively.


andrewgodwin (approved):
Ah right, I must have broken this when I fixed usernames to preserve case, as we found a server whose actor URI was case sensitive!

PR #266, opened 2022-12-25
Quality of life

tabletcorry:
A few quality of life fixes picked up as I set up my development environment for the first time, and created my first python patch. Nothing too exciting here.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -43,6 +43,8 @@ services:
       POSTGRES_DB: takahe
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: insecure_password
+    ports:
+      - "5432:5432"
 
   web:
     <<: *takahe-common
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -44,10 +44,12 @@ services:
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: insecure_password
     ports:
-      - "5432:5432"
+      - "5433:5432"
 
   web:
     <<: *takahe-common
+    environment:
+      GUNICORN_EXTRA_CMD_ARGS: "--reload"
     ports:
       - "8000:8000"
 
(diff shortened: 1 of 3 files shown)


docker/run.sh
```diff
@@ -4,10 +4,14 @@
...
 
+# Enable auto-reload when in DEBUG mode
+[ "${TAKAHE_DEBUG}" = "true" ] && GUNICORN_RELOAD="--reload"
+
 # Run nginx and gunicorn
 nginx -c "/takahe/docker/nginx.conf" &
 
-gunicorn takahe.wsgi:application -b 0.0.0.0:8001 &
+echo $GUNICORN_RELOAD 4
+gunicorn takahe.wsgi:application -b 0.0.0.0:8001 $GUNICORN_RELOAD &
```
manfre, line 14: `GUNICORN_CMD_ARGS` exists to override the options

docker/docker-compose.yml
```diff
@@ -43,6 +43,8 @@ services:
       POSTGRES_DB: takahe
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: insecure_password
+    ports:
+      - "5432:5432"
```
TkTech, line 47: Keep this in your local version please, don't need port conflicts when postgres is running on the hos.

docker/run.sh
TkTech: We should probably make it the default in the docker-compose for dev

tabletcorry: I am hesitant to use `GUNICORN_CMD_ARGS` since it is already in use by the Dockerfile and likely to be the first thing people reach for (I use it for configuring my deployment environment on fly.io as well).

So I will set a new env var in docker compose for this specific purpose. Commit incoming.

docker/docker-compose.yml
tabletcorry: How about I switch it to `5433`? That port is pretty much unused.

 It is quite useful to have this exposed as a default so I don't need to modify a in-tree file all the time.

andrewgodwin: It's going to clash for someone I'm sure, but let's stick it in there and see if anyone complains.

andrewgodwin:
Thanks!

PR #267, opened 2022-12-25
Basic OpenGraph support

tabletcorry:
Creates an OpenGraph template include in base.html including the basic tags expected on all pages.

Then allows any page to add additional expected tags via `context`.

Currently, profiles and posts are enriched to show complete opengraph metadata, and render correctly in Discord.

Note: This does not show posts in Slack like Twitter/Mastodon do. I believe this is due to Slack preferring oembed when present, which is a mastodon API endpoint we may need to create at some point.

### DIFF
--- a/activities/views/posts.py
+++ b/activities/views/posts.py
@@ -42,8 +42,6 @@ class Individual(TemplateView):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
-        context["opengraph"] = self.post_obj.to_opengraph_dict()
-
         ancestors, descendants = PostService(self.post_obj).context(
             self.request.identity
         )
--- a/activities/views/posts.py
+++ b/activities/views/posts.py
@@ -57,13 +57,19 @@ class Individual(TemplateView):
             "opengraph": {
                 "title": f"{self.post_obj.author.name} (@{self.post_obj.author.handle})",
                 "type": "article",
-                "published_time": self.post_obj.published.isoformat()
-                or self.post_obj.created.isoformat(),
-                "modified_time": self.post_obj.edited.isoformat()
-                or self.post_obj.published.isoformat()
-                or self.post_obj.created.isoformat(),
-                "description": self.post_obj.summary
-                or bleach.clean(self.post_obj.safe_content_local(), strip=True),
+                "published_time": (
+                    self.post_obj.published.isoformat()
+                    or self.post_obj.created.isoformat()
+                ),
+                "modified_time": (
+                    self.post_obj.edited.isoformat()
+                    or self.post_obj.published.isoformat()
+                    or self.post_obj.created.isoformat()
+                ),
+                "description": (
+                    self.post_obj.summary
+                    or bleach.clean(self.post_obj.safe_content_local(), strip=True)
+                ),
                 "image": {
                     "url": self.post_obj.author.local_icon_url().absolute,
                     "height": 85,
--- a/activities/views/posts.py
+++ b/activities/views/posts.py
@@ -1,4 +1,3 @@
-import bleach
 from django.core.exceptions import PermissionDenied
 from django.http import Http404, JsonResponse
 from django.shortcuts import get_object_or_404, redirect, render
@@ -67,8 +66,7 @@ class Individual(TemplateView):
                     or self.post_obj.created.isoformat()
                 ),
                 "description": (
-                    self.post_obj.summary
-                    or bleach.clean(self.post_obj.safe_content_local(), strip=True)
+                    self.post_obj.summary or self.post_obj.safe_content_local()
                 ),
                 "image": {
                     "url": self.post_obj.author.local_icon_url().absolute,
--- a/activities/views/posts.py
+++ b/activities/views/posts.py
@@ -1,3 +1,4 @@
+import bleach
 from django.core.exceptions import PermissionDenied
 from django.http import Http404, JsonResponse
 from django.shortcuts import get_object_or_404, redirect, render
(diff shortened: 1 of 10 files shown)


tabletcorry:
Attaching a few samples of the link expansion. Discord looks great, and Slack looks... ok.

Screenshot 2022-12-24 at 7 40 49 PM
Screenshot 2022-12-24 at 7 40 39 PM
Screenshot 2022-12-24 at 7 40 35 PM

activities/views/posts.py
```diff
@@ -53,6 +54,22 @@ def get_context_data(self):
...
+            "opengraph": {
+                "title": f"{self.post_obj.author.name} (@{self.post_obj.author.handle})",
+                "type": "article",
+                "published_time": self.post_obj.published.isoformat()
+                or self.post_obj.created.isoformat(),
+                "modified_time": self.post_obj.edited.isoformat()
+                or self.post_obj.published.isoformat()
+                or self.post_obj.created.isoformat(),
+                "description": self.post_obj.summary
+                or bleach.clean(self.post_obj.safe_content_local(), strip=True),
```
andrewgodwin, line 66: There's a `core.html.strip_html` function I'd rather you use here.

activities/views/posts.py
```diff
@@ -53,6 +54,22 @@ def get_context_data(self):
...
             "ancestors": ancestors,
             "descendants": descendants,
+            "opengraph": {
+                "title": f"{self.post_obj.author.name} (@{self.post_obj.author.handle})",
+                "type": "article",
+                "published_time": self.post_obj.published.isoformat()
+                or self.post_obj.created.isoformat(),
+                "modified_time": self.post_obj.edited.isoformat()
+                or self.post_obj.published.isoformat()
+                or self.post_obj.created.isoformat(),
```
andrewgodwin, line 64: If you wrap these `or` sets in parentheses, the auto-formatter will indent them, which I think is more readable.

activities/views/posts.py
tabletcorry: I looked at that, but it linkifies, which we can't accept here. OpenGraph requires bare text or it rejects AFAICT.

Should I put a boolean param on that function?

activities/views/posts.py
tabletcorry: Ah, ok. Will do!

activities/views/posts.py
andrewgodwin: Hmm, you could, or you could consider `html_to_plaintext` in that same module which will also turn <p> and <br> back into newlines if you wanted that.

templates/_opengraph.html
```diff
@@ -0,0 +1,32 @@
...
+        <meta content="{{ opengraph.description }}" property="og:description"/>
+    {% endif %}
+    {% if opengraph.profile.username %}
+        <meta content="{{ opengraph.profile.username }}" property="og:profile:username"/>
+    {% endif %}
+    <meta content="{{ opengraph.image.url|default:config.site_icon }}" property="og:image"/>
+    <meta content="{{ opengraph.image.type|default:"image/webp" }}" property="og:image:type"/>
+    {% if opengraph.image.width %}
+        <meta content="{{ opengraph.image.width }}" property="og:image:width" />
+        <meta content="{{ opengraph.image.height }}" property="og:image:height" />
```
phildini, line 26: When I compare this to what's on mastodon, here's what I think is missing from meta tags that's worth adding because we don't know how legacy systems are parsing tags:

- `name='description'` -- should probably match `og:description`
- `property='og:image:alt'` -- if we don't have alt tags today, maybe add the property and add an xfail test so we remember?

activities/views/posts.py
tabletcorry: Ah, missed that. But since we switched to `|striptags` for now this may be mooted

templates/_opengraph.html
tabletcorry: Added both, and haven't done tests yet since the code is still being debated on best practices :)

templates/base.html
```diff
@@ -25,6 +25,7 @@
     {% if config_identity.custom_css %}
       <style>{{ config_identity.custom_css }}</style>
     {% endif %}
+    {% include "_opengraph.html" %}
```
manfre, line 28: Use a block here with the generic opengraph contained within. E.g.
```
{% block opengraph %}
    {% include "_opengraph.html" og_data=opengraph %}
{% endblock %}
```
Pages that need to disable opengraph would include an empty block `{% block opengraph %}{% endblock %}`.

Actor pages would mimic the base's block above with `og_data=post.to_opengraph_dict`.

tabletcorry: Nod, I more or less did that, but inside of `_opengraph.html` (so I wouldn't clutter base as much). Do you prefer the block to be in base rather than the include?

activities/views/posts.py
```diff
@@ -39,42 +39,30 @@ def get(self, request, handle, post_id):
             # Show normal page
             return super().get(request)
 
-    def get_context_data(self):
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        context["opengraph"] = self.post_obj.to_opengraph_dict()
```
manfre, line 45: This is not needed, `post.to_opengraph_dict` is usable within the template.

templates/base.html
manfre: Change this to the example I put above:
```
{% block opengraph %}
    {% include "_opengraph.html" og_data=opengraph %}
{% endblock %}
```

In `templates/identity/view.html`:
```
{% block opengraph %}
    {% include "_opengraph.html" og_data=identity.to_opengraph_dict %}
{% endblock %}
```

In `templates/activities/post.html`:
```
{% block opengraph %}
    {% include "_opengraph.html" og_data=post.to_opengraph_dict %}
{% endblock %}
```

manfre (commented):
Added comments about how this can be handled in the templates.

activities/views/posts.py
tabletcorry: Yes, but I don't know that I need to call `post` or `identity`, or some future object, or the detault OG keyset, so I need some kind of indirection.

tabletcorry: Oh, I see your other comment. Looking.

andrewgodwin (approved):

andrewgodwin:
Thanks!

PR #269, opened 2022-12-25
Fallback avatars in about 500 bytes.

TkTech:
Adds a bit of Javascript, but it's 100% optional and works fine if it's missing or blocked.

Generates consistent avatars for a given handle, doing about 100 in 1ms.

You may want to convert the `onerror` handle to hyperscript, I'm simply not quite sure how onerror is supposed to work with it.

image


static/js/minidenticons.min.js
```diff
@@ -0,0 +1,8 @@
+function identicon(t,e=50,i=50){const n=t.split("").reduce(((t,e)=>16777619*((t^e.charCodeAt(0))>>>0)),2166136261);const s=n/16777619%18*20;return[...Array(t?25:0)].reduce(((t,e,i)=>n%(16-i%15)<4?t+`<rect x="${i>14?7-~~(i/5):~~(i/5)}" y="${i%5}" width="1" height="1"/>`:t),`<svg viewBox="-1.5 -1.5 8 8" xmlns="http://www.w3.org/2000/svg" fill="hsl(${s} ${e}% ${i}%)">`)+"</svg>"}
+function generate_avatar(element) {
+  element.src = URL.createObjectURL(
+    new Blob([identicon(element.dataset.handle)], {
+      type: 'image/svg+xml;charset=utf8'
+    }
+  ));
```
manfre, line 7: Replace `generate_avatar` with:
```
function generate_avatar(handle) {
    return URL.createObjectURL(new Blob([identicon(handle)], {type: 'image/svg+xml;charset=utf8'}));
}
```

and use it on the images with:
```
_="on error set my.src to generate_avatar(@data-handle)"
```

TkTech:
I'd pull this one to give it a try before merging but it should be good to go.

andrewgodwin (approved):

PR #273, opened 2022-12-26
Render the Nginx config to a separate file

tabletcorry:
When writing the cache size to nginx config, write it to a separate file so it doesn't confuse git status/commit.

By using a rendered file, we can ignore the output completely.


andrewgodwin:
Yes, I guess if you mount the takahe dir this is needed - thanks!

PR #274, opened 2022-12-26
Move  nginx to separate container

tabletcorry:
By moving nginx to its own container, we make it _vastly_ easier to insert a debugger into the environment. Once gunicorn/django is on its own container, we can inject our debugger there without affecting the nginx process.

This assumes that nginx and gunicorn don't have any shared files on disk, which doesn't appear to be the case today. If such a thing is added, a shared volume will be required.

Specifically, this unlocks trivial PyCharm debugging (and presumably other debuggers)

Additionally:
* Create dependency tree within all containers, and specify healthcheck should complete
* Expose all container service ports for debugging
* Improve healthcheck timing on postgres container

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -24,8 +24,6 @@ x-takahe-common:
       - external_network
       - internal_network
     restart: on-failure
-    depends_on:
-      - db
     volumes:
       - ..:/takahe/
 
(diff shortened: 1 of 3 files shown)


andrewgodwin:
I'm afraid I am not willing to do this to the main image. I want a single webserver image where everything is configured within it.

If you want to find a way to do debugging that doesn't affect the way the image runs I'm open to it, but the production image should contain everything.

tabletcorry:
Sounds good, will revise.

The image does still contain everything (nginx and web use the same container), but I did modify a few config files in a way that makes it slightly incompatible.

Should be able to fix that though.

andrewgodwin:
Yeah, sorry, I more meant "they should all run at once". My main design goal for the Docker container is production deployment being relatively simple; development environment priorities matter, but they're second in line and shouldn't affect the main deployment.

Issue #248, opened 2022-12-24
tooting from toot! - can't make toots


whatSocks:
see error 

![IMG_0ED623DC5C89-1](https://user-images.githubusercontent.com/2731744/209419309-75504361-887b-4c5b-9024-5790c02ec64a.jpeg)

andrewgodwin:
Fascinating. We'll have to debug this with it connected against a live test server, unfortunately, which means I need to go dig out an ipad at some point.

whatSocks:
this is fixed

PR #275, opened 2022-12-26
Fixes for the Toot! client (#248)

TkTech:
This is a WIP, debugging with the author of Toot!.

- Truncates timestamps returned by the API to 3 digits from 6, to be identical with Mastodon.
- Adds a setting to disable the public timeline and moves timeline-related settings into a Timeline settings group.
  - Also allows the public timeline without requiring an identity by default because, well, it's the public timeline.
- Use absolute URLs for hashtags in posts, just like we do with AP serialization.


st3fan:
Does this patch also address the following error?

> The data couldn’t be read because it is missing. (Expected URL value but found null instead) header

I think this is a native error that the app is presenting after going through to auth flow of Takahe 0.6 in Toot. 

(The “header” part at the end is a but unclear. May be a message formatting error.)

TkTech:
This PR will eventually address all issues with the toot! client to get to a MVP (login, see your timeline, post a "toot").

TkTech:
> Does this patch also address the following error?
> 
> > The data couldn’t be read because it is missing. (Expected URL value but found null instead) header
> 
> I think this is a native error that the app is presenting after going through to auth flow of Takahe 0.6 in Toot.
> 
> (The “header” part at the end is a but unclear. May be a message formatting error.)

Your specific issue is resolved in [77c2a2e](https://github.com/jointakahe/takahe/pull/275/commits/77c2a2e5e0d9512dddaf73b5ad2b4cf0fee4e464), however there are still several other issues to resolve.

andrewgodwin (approved):

PR #276, opened 2022-12-26
Add docker healthcheck feature & move collectstatic into container

kromonos:
Install netcat for docker healthcheck feature
Move collectstats to docker setup container


andrewgodwin (changes requested):
The collectstatic command has to be in the container so that it is in the final published image. You can put it in the setup container as well if you want, but it cannot be removed from the main build.

kromonos:
> The collectstatic command has to be in the container so that it is in the final published image. You can put it in the setup container as well if you want, but it cannot be removed from the main build.

oh ok. I've reverted that change.

docker/docker-compose.yml
```diff
@@ -49,6 +49,11 @@ services:
 
   web:
     <<: *takahe-common
+    healthcheck:
+      test: ["CMD", "nc", "-z", "-v", "localhost", "8000"]
```
manfre, line 53: Is this health check essentially "connect to the port, but don't do anything with the connection"? If so, curl to `/actor/` would let us know that both nginx & takahe are running well enough to return a vital (fast) endpoint.

kromonos: It tells docker, that the service is operating healthy or not. With `docker ps` it would look like this:  


But yes, one could connect to nginx too, but I thought I keep it out for the moment because of https://github.com/jointakahe/takahe/pull/278 which splits some things up.

andrewgodwin (approved):

andrewgodwin:
This is good for now - we can revisit the healthcheck if the split does happen!

PR #278, opened 2022-12-26
Introduce debugger friendly docker-compose-split.yml

tabletcorry:
To support the new debugger compose definition, tweak run.sh to spawn nginx/gunicorn from function calls with kill switches. We can then call these kill switches as needed during development mode.

This change has no impact on production or default workflows.

### DIFF
--- a/docs/ide.md
+++ b/docs/ide.md
@@ -1,0 +1,29 @@
+# IDE Development
+
+IDE development and debugging should be possible, and is currently tested/used in PyCharm by some developers.
+
+Please document additional IDEs below if you are familiar with how to set them up.
+
+## PyCharm
+
+To configure full development/debugging in PyCharm, please run through the checklist below:
+
+1. Create a python interpreter within Docker Compose
+    1. Go to Settings > Project > Python Interpreter
+    2. Add Interpreter > Docker Compose
+        * Server: Docker
+        * Configuration Files: `docker/docker-compose-split.yml`
+        * Service: `app`
+    3. Click next and select the auto-populated system interpreter (`/usr/local/bin/python3`)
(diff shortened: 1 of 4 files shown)


docs/ide.md
```diff
@@ -0,0 +1,29 @@
+# IDE Development
+
```
pauloxnet, line 2: This addition seems off-topic. I suggest to open a separated PR for IDE support.

tabletcorry: The PR is specifically to enable IDE development, so I documented how to use my changes. It could be a separate PR I suppose, but that seems a little too broken out for my usual tastes 🤔

pauloxnet: Sorry but the IDE part was not obvious from the title

tabletcorry:
No rush, but do you have any thoughts on this @andrewgodwin? I think this is sufficiently out of the way that no typical user will notice the change.

andrewgodwin:
This fiddles with run.sh a bit too much for my liking - I'd rather not gain any more complexity in there. Is there literally no other way to get the debugger enabled?

tabletcorry:
I need some way to turn off gunicorn and nginx. The run.sh change can be structured any way that creates a switch for each. I used functions only because they looked a bit more organized.

If you strongly object to any change there, I can make clones of run.sh, but I wanted to try and keep it unified.

andrewgodwin:
Sorry I've let this sit for so long! Given we also have the ability to run Takahē directly from a virtualenv locally, would that not make more sense for debugging? The new makefile commands in #358 will make database handling in that situation reasonable too.

andrewgodwin:
Closing this due to inactivity - feel free to reopen if you have more progress/feedback.

PR #280, opened 2022-12-27
Identity profile metadata input with dynamic form

manfre:
It works, but could likely use style and template cleanup. Leaving as draft since I'm not sure how much more I'm going to tinker with this. I might abstract this out to a generic key/value input _field and make another for single value (will be needed for also_known_as). The hyperscript to make a generic multivaluefield is beyond me at this point.

image

image

### DIFF
--- a/templates/settings/profile.html
+++ b/templates/settings/profile.html
@@ -3,15 +3,19 @@
 {% block subtitle %}Profile{% endblock %}
 
 {% block content %}
-    <form action="." method="POST" enctype="multipart/form-data" >
+    <form action="." method="POST" enctype="multipart/form-data"
+        _="on submit metadata.collectMetadataFields()">
         {% csrf_token %}
+
         <fieldset>
             <legend>Details</legend>
             {% include "forms/_field.html" with field=form.name %}
             {% include "forms/_field.html" with field=form.summary %}
+            {% include "forms/_json_name_value_list.html" with field=form.metadata %}
             {% include "forms/_field.html" with field=form.discoverable %}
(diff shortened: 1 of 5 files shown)


templates/settings/profile.html
```diff
@@ -3,15 +3,19 @@
 {% block subtitle %}Profile{% endblock %}
 
 {% block content %}
-    <form action="." method="POST" enctype="multipart/form-data" >
+    <form action="." method="POST" enctype="multipart/form-data"
+        _="on submit metadata.collectMetadataFields()">
```
manfre, line 7: I haven't figured out how to avoid needing to add this to the form `on submit`.

manfre:
image

andrewgodwin (commented):
So, this looks fine, but one question - do we need a dynamic form? Mastodon just has four of these ready to go all the time, and it would make it simpler to maintain if we just did that.

manfre:
Having 4 hardcoded fields is an option, but I went dynamic for a few reasons:
* our page layout is vertically longer than Mastodon's. Have 3-4 rows of empty inputs doesn't help.
* I want more than 4 links. This allows a profile page to double as a "Little Link" page.

andrewgodwin:
Have you checked what happens when you send more than 4 links over to Mastodon? I just want to make sure we don't break them.

manfre:
Yes, it shows them.

image

andrewgodwin:
Neat!

andrewgodwin (approved):

andrewgodwin:
This looks a little weird when it's empty and has no fields - is there a way we can start with 2 empty ones?

PR #281, opened 2022-12-27
Add cache-control header for static assets

tabletcorry:
Easy to do this for the static assets, since they have cache-busting hashes.

I think we may want to move the proxy images to have hash filenames as well, like Mastodon does, so we can set the `immutable` cache setting on them as well. Something for later though.

### DIFF
--- a/docker/nginx.conf
+++ b/docker/nginx.conf
@@ -38,6 +38,9 @@ http {
 
         # Serves static files from the collected dir
         location /static/ {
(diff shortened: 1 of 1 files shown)


andrewgodwin (approved):

andrewgodwin:
Thanks!

The proxied assets are harder - Mastodon actually fully downloads them to its normal media system rather than proxy-and-cache, so they can use their normal naming. I don't want to do that because it results in a lot of server admin cleaning out that storage directory continuously.

PR #288, opened 2022-12-27
More aggressive caching

rcarmo:
This is a quick fix to ensure assets are cached upon the very first request


andrewgodwin:
Thanks!

PR #292, opened 2022-12-27
Add ability to customise Django settings

manfre:
@andrewgodwin PR is intended as a decision point about how we'd like to encourage people to handle custom settings overrides.

1. People set `DJANGO_SETTINGS_MODULE` and then do whatever they're going to do, optionally importing our settings.py
    * More flexible
    * Harder to troubleshoot and document
2. This PR, we do the normal settings.py before optionally importing a fixed path
    * Less flexible (requires `DJANGO_SETTINGS_MODULE` for any pre-Takahē settings code)
    * Common pattern that can be documented
    * Solves the common case of "add to INSTALLED_APPS" and add extra settings without people potentially not importing settings.py

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 /docs/_build
 /media/
 /static-collected
+/takahe/local_settings.py
 __pycache__/
 api-test.*
 fly.toml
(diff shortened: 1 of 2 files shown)


andrewgodwin:
I prefer the second option that this PR presents, but I'd like to call it `local_settings.py` due to personal preference!

andrewgodwin (approved):

Issue #287, opened 2022-12-27
[Suggestion] extensions in cacheable media


rcarmo:
Hi there, 

I'm trying to tune caching in Takahe, and have noticed that media/proxy URLs don't have a file extension. Since I use Cloudflare, having an extension [would actually simplify edge caching](https://developers.cloudflare.com/cache/about/default-cache-behavior). 

That can be easily verified by running a public URL via https://redbot.org.

For instance, using my test instance (which is behind Cloudflare and I've been tuning `nginx` caching on):

* [This uploaded PNG file is cached by Cloudflare](https://redbot.org/?uri=https://insightful.systems/media/profile_images/2022/12/20/BQvqy_viXWmdB4leJvlHludq5Ns.png)

```
HTTP/1.1 200 OK
Date: Tue, 27 Dec 2022 13:54:13 GMT
Content-Type: image/png
Content-Length: 27432
Connection: keep-alive
Cache-Control: public, max-age=14400
Content-Disposition: inline;
        filename="BQvqy_viXWmdB4leJvlHludq5Ns.png
        4;
Cross-Origin-Opener-Policy: same-origin
Last-Modified: Tue, 20 Dec 2022 23:54:47 GMT
Referrer-Policy: same-origin
Vary: Cookie, Origin
X-Cache: MISS
X-Content-Type-Options: nosniff
X-Frame-Options: DENY
CF-Cache-Status: HIT
Age: 206
Accept-Ranges: bytes
Report-To: {"endpoints":[{"url":"https:\/\/a.nel.cloudfl
        are.com\/report\/v3?s=wkjpQCMvFt3uHOv%2Bvm4UQk25QLyn0uNeOa
        i5i0TymG9b9WgTThlzuEOI6LwARUePPjkuYNp%2B1m%2BvHeJ1XzQDY3hp
        eUCBpltP%2BqvPubX5y5a3dYqISQjpl0xY4DXdz%2B6Sd85siVsfiL54LT
        OuFr7cgAI%3D"}],"group":"cf-
        nel","max_age":604800}
NEL: {"success_fraction":0,"report_to":"cf-
        nel","max_age":604800}
Server: cloudflare
CF-RAY: 78028c0348e597cf-SJC
alt-svc: h3=":443"; ma=86400, h3-29=":443"; ma=86400
```

* [This proxied banner image is not cached by Cloudflare](https://redbot.org/?uri=https://insightful.systems/proxy/identity_image/27/)

```
[...]

PR #294, opened 2022-12-27
Use cache-busting URLs for proxied files

tabletcorry:
Migrates (in a backwards-compatible way) from `/proxy/identity_image/271/` to `/proxy/identity_image/271/f5d8e72f2b/`.

The "hash tail" is generated as a truncated SHA1 of the upstream URL. This allows us to respect the `Cache-Control` header we get from upstream, as if the image changes for the identity then the hash will also change.

In other words, while the previous arrangement forced Takahē to replace the cache-headers because it didn't know when upstream changed, the new setup will immediately update the URL when the upstream does.

Overall, this gives us two benefits:
* Upstream cache settings are preserved, respecting the server's wishes
* Any upstream change will immediately propagate through all caches, including nginx and any CDNs

Nothing actually parses the hash, it exists only to bust the cache. The security of the hash doesn't matter, and the length doesn't really matter either (because it is scoped to the identity PK).

Any attempts to reach the "old" URL without a hash will continue to work, though the upstream cache-control headers will be used.

Note: If an upstream does have "bad" cache control settings, this will allow it to affect the CDN/client. While not ideal, this seems a reasonable tradeoff (and nginx will maintain its cache independently).

To see this in action, it is deployed to my test server: https://takahe.tabletcorry.com/@mark@tapbots.social/

### DIFF
--- a/activities/models/emoji.py
+++ b/activities/models/emoji.py
@@ -168,7 +168,9 @@ class Emoji(StatorModel):
             if self.file:
                 return AutoAbsoluteUrl(self.file.url)
             elif self.remote_url:
-                return AutoAbsoluteUrl(f"/proxy/emoji/{self.pk}/")
+                return AutoAbsoluteUrl(
+                    f"/proxy/emoji/{self.pk}/", hash_tail_input=self.remote_url
+                )
         return StaticAbsoluteUrl("img/blank-emoji-128.png")
 
     def as_html(self):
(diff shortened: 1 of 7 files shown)


andrewgodwin:
While I agree with the idea, I'm not entirely psyched with having to break the URL abstraction by passing in the remote URL to do it; did you investigate trying to have nginx do this instead?

tabletcorry:
I didn't think to do it in nginx, no. Since Django handles the proxy in the degenerate case, it seems like it should be there.

It also makes the system more resilient to technology changes in the nginx layer, since that may change at some point.

andrewgodwin:
OK, having slept on it, I agree with you, but I'm also going to modify this after I pull it to satisfy #287 and put extensions on the end, if we're doing this anyway.

PR #295, opened 2022-12-27
Include Identity metadata fields in ActivityPub messages

manfre:
Related to #280 PR, but can be merged independently.

### DIFF
--- a/core/html.py
+++ b/core/html.py
@@ -45,14 +45,14 @@ def sanitize_html(post_html: str) -> str:
     return mark_safe(cleaner.clean(post_html))
 
 
-def strip_html(post_html: str) -> str:
+def strip_html(post_html: str, *, linkify: bool = True) -> str:
     """
     Strips all tags from the text, then linkifies it.
     """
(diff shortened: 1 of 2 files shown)


andrewgodwin (approved):

core/ld.py
```diff
@@ -429,6 +429,8 @@ def canonicalise(json_data: dict, include_security: bool = False) -> dict:
             "sensitive": "as:sensitive",
             "toot": "http://joinmastodon.org/ns#",
             "votersCount": "toot:votersCount",
+            "schema": "http://schema.org#",
+            "PropertyValue": "schema:PropertyValue",
```
andrewgodwin, line 433: If you do this, you also need to change everywhere else that uses `http://schema.org#PropertyValue` as a lookup value (e.g. `fetch_actor`)

andrewgodwin (changes requested):

core/ld.py
manfre: Rather inline it.

andrewgodwin (approved):

PR #298, opened 2022-12-28
Guess emoji mimetype if not provided in tag

manfre:
Some servers don't send `mediaType` or bury it using `und` (for some odd reason). If it's not found as a singular value, we'll try to infer the type from the url.

### DIFF
--- a/activities/models/emoji.py
+++ b/activities/models/emoji.py
@@ -1,3 +1,4 @@
+import mimetypes
 import re
 from functools import partial
 from typing import ClassVar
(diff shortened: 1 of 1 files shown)


TkTech:
I've added libmagic as a dependency in the arbitrary uploads PR, it can be used for this in the future to detect the file type.

manfre:
@TkTech it can't be used here because libmagic requires the file, but we can correct the mimetype when we download it in `handle_outdated`

andrewgodwin (approved):

PR #310, opened 2022-12-29
SECRET_KEY prod guard breaks Dockerfile

manfre:


andrewgodwin:
I'd rather we set a useless secret key rather than turn on debug during static collection I think?

manfre:
Didn't notice the other PR. Setting a static secret works too.

PR #311, opened 2022-12-29
Create Django stubs aware mypy config

tabletcorry:
The script is a bit of a hack, so open to suggestions. It does work, though the output is rather noisy. Currently, this is useful for occasional manual review of uncaught typing issues.

May be better to cram this in a folder, but wasn't sure what to name it. Worst case, this may help someone in the future who knows how to get this cleanly into pre-commit.

The main issue for pre-commit is that you need _all_ the packages installed for this version of mypy to work, and pre-commit doesn't have them installed in the environment. In theory pre-commit could call this script, but it will be a bit slow/messy.

### DIFF
--- a/.mypy_django.ini
+++ b/.mypy_django.ini
@@ -1,0 +1,5 @@
+[mypy]
+plugins = mypy_django_plugin.main
+
+[mypy.plugins.django-stubs]
+django_settings_module = "takahe.settings"
--- a/mypy_django.sh
+++ b/mypy_django.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+
+set -o errexit
+
+# Change directory to the project root directory.
+cd "$(dirname "$0")"
+
+# The HEAD version of django-stubs is required due to unreleased fixes
+pip install --quiet --no-input \
+  -r requirements-dev.txt \
+  types-pyopenssl \
+  types-bleach \
+  types-mock \
+  types-cachetools \
+  types-python-dateutil \
+  types-docutils \
+  mypy \
+  git+https://github.com/typeddjango/django-stubs.git@master
+
+export TAKAHE_DATABASE_SERVER="postgres://x@example.com/x"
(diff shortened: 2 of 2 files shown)


andrewgodwin:
So elaborate further why this can't be in pre-commit? Is it the HEAD version of the stubs needed?

I'm loathe to commit something like this that will just not get run by anyone rather than a potentially more limited version that will actually integrate into pre-commit.

tabletcorry:
Agreed, I don't love having it stand apart, so if ultimately this gets closed as a note for future reference, that is fine by me.

I will expand on the pre-commit issues in the morning 👍

tabletcorry:
Ok, so here were the various findings:

(a) If you try to directly enable Django-stubs in pre-commit (e.g. by adding `Django-stubs` to `additional_dependencies`), it will fail immediately because it needs environment variables (which pre-commit doesn't really support readily) and all of the packages in `requirements.txt` which pre-commit just won't install (because the requirements file isn't present at dependency install time).

(b) An option is to use the script I created inside of pre-commit, which is discussed in link [1]. I considered doing this, but the output of Django-aware mypy is _very noisy_ and I am not sure it is possible to clean it up.

I think the main next step is to determine if the output can be reduced to a quiet output. My other PR knocks out about 10 errors, but there are another 60 errors or so in the complete output, and some may be due to incomplete stubs.

[1]: https://jaredkhan.com/blog/mypy-pre-commit

andrewgodwin:
Aha, got it. Given those don't seem surmountable easily, maybe the move is to make this into something that can either be run as one of our `make` commands (I'd suggest `make typecheck`) and then we can also run it in the CI to enforce it?

tabletcorry:
Nod, that's a good idea. Will take a look at making my script a bit more CI/CD appropriate.

andrewgodwin:
Closing this out since we've not got back to it in months - can be reopened if needed.

PR #312, opened 2022-12-29
Fixes developed from mypy_django script

tabletcorry:
A few discoveries from running mypy with Django-awareness. Nothing critical, but it caught a few odds and ends.

### DIFF
--- a/core/files.py
+++ b/core/files.py
@@ -2,10 +2,15 @@ import io
 
 import blurhash
 import httpx
+import PIL.ImageQt
 from django.conf import settings
 from django.core.files import File
 from django.core.files.base import ContentFile
 from PIL import Image, ImageOps
+
+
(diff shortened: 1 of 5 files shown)


core/files.py
```diff
@@ -2,19 +2,24 @@
 
 import blurhash
 import httpx
+import PIL.ImageQt
```
andrewgodwin, line 5: Wait what is this doing here, since it's not referenced elsewhere?

tabletcorry: Uh, good question. I have no idea.

tabletcorry: I think something misfired when I typed it, and imported the reference. Anyways, fixed now.

andrewgodwin (approved):

andrewgodwin:
Lovely, thanks!

PR #317, opened 2022-12-29
Option to show/hide boosts for a followed user

cosmin:
Support for hiding someone while hiding their boosts. The AP schema refers to boosts as reblogs, as does the client API, so for consistency I also called the field reblogs on the follow schema.

Currently this only implements this on the schema and in the API, no UI is exposed to show/hide boosts on a follow.


users/services/identity.py
```diff
@@ -75,7 +82,7 @@ def mastodon_json_relationship(self, from_identity: Identity):
             "requested": False,
             "domain_blocking": False,
             "endorsed": False,
-            "note": "",
+            "note": follow.note or "",
```
cosmin, line 85: this is unrelated to show/hide boosts but it's another property of the follow object which was previously unavailable. I can split this into a separate PR if that would be cleaner.

andrewgodwin (changes requested):
I know this is annoying, but I'm going to request that it is called "boosts" not "reblogs" for internal codebase consistency

users/services/identity.py
andrewgodwin: This is fine!

andrewgodwin (approved):
Looks good, but I'm going to wait until after 0.6.1 is released to merge this as it has a migration.

PR #318, opened 2022-12-30
Identity profile menus

manfre:
This changes the "Select Identity" link in the header to a profile link for the current identity. "Select Identity" link has been moved to _menu.

Adds an admin drop down menu to all profiles and relocates the menu action form for (un)follow to only surround those actions to allow for future actions.

image

image

image

### DIFF
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -285,6 +285,11 @@ header menu a.identity i {
     padding: 0 7px 2px 0;
 }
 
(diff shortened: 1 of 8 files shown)


andrewgodwin (approved):

andrewgodwin:
Thanks! We can consider making the top-right thing have a little menu arrow of identities itself at some future point.

PR #319, opened 2022-12-30
/v1/accounts/{id}/following and /v1/accounts/lookup support.

TkTech:
Fixes for https://followgraph.vercel.app/ functionality as reported by @cosmin.

### DIFF
--- a/api/views/accounts.py
+++ b/api/views/accounts.py
@@ -1,3 +1,4 @@
+from django.db.models import Q
 from django.http import HttpRequest, HttpResponse, JsonResponse
 from django.shortcuts import get_object_or_404
 from ninja import Field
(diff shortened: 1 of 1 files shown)


api/views/accounts.py
```diff
@@ -171,3 +193,40 @@ def account_unfollow(request, id: str):
...
+    )
+
+    if not identity.config_identity.visible_follows and request.identity != identity:
+        return []
+
+    service = IdentityService(identity)
+
+    paginator = MastodonPaginator(Identity)
+    pager = paginator.paginate(
+        service.following(),
```
andrewgodwin, line 219: I don't believe this will paginate properly as the queryset is ordered by name but the paginator is left on its default sort attribute of created

andrewgodwin (approved):

PR #321, opened 2022-12-30
A quick robots.txt to tell bots to stop crawling tags.

TkTech:
One possible approach to add a simple robots.txt to stop bots that are aggressively crawling the tags.


andrewgodwin (commented):
Looks good, but you need a final newline to satisfy the linter!

andrewgodwin (approved):

PR #332, opened 2022-12-31
Add indexes to help Stator and viewing posts

manfre:


andrewgodwin (commented):
Hmm, is there no way we can get these on the base `StatorModel` so that we don't forget in future?

andrewgodwin (approved):
Lovely, it's nicer when it's inherited, even if we do have to play the weird subclass tricks.

PR #334, opened 2022-12-31
Make accessibility text color visible

shuuji3:
This is a small tweak to change the navigation text color to white. These links are useful for keyboard users too. I believe making visible doesn't affect usual users as these are only shown when fucusing.

Screenshot 2023-01-01 at 1 38 20
Screenshot 2023-01-01 at 1 38 09

Here are the similar real world examples:
- [ARIA: navigation role - Accessibility | MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role)
- [Guidance on Web Accessibility and the ADA | ADA.gov](https://www.ada.gov/resources/web-guidance/)

### DIFF
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -1736,7 +1736,7 @@ form .post {
 }
 
 .screenreader-text:focus {
-    color: black;
+    color: var(--color-text-main);
     display: inline-block;
     height: auto;
     width: auto;


andrewgodwin (approved):

andrewgodwin:
Thanks!

PR #338, opened 2023-01-01
oauth2 fixes

cosmin:
This implements a few oauth2 fixes:

- passes along the state object
- enforces authorization code expiration (currently set to 1 minute, we could make this configurable)
- enforces redirect_uri
- properly checks for client_secret when granting a token
- handles pulling client authentication for token grant from basic auth
- implement token revocation

In order to implement expiration I split out the authorization from the token into a separate table. Since these two have different lifetimes, with the authorization valid for a short period of time, and the token valid for a long period of time, keeping them in one model makes it harder to reason about. 

One time use of authorization is enforced by checking whether there is an associated token object.


cosmin:
I tested this with Metatext and Tooot and was able to log in and log out with both, and verified that tokens are properly revoked

andrewgodwin:
Just so I'm clear - is the authorization expiry the only thing making a separate table required?

cosmin:
I wouldn't say required, it's certainly possible to beat both of these concepts into the same table, but I do think that it's better to keep them separate primarily because of the different lifetimes. We also need to store redirect_uri with the authorization to verify on the token grant. Since authorizations have a short lifetime we could clean up all of this periodically.

It's also the common pattern for implementing this (unless using stateless authorization codes which encode all of the information in the code itself). For example https://django-oauth-toolkit.readthedocs.io/en/latest/models.html#oauth2_provider.models.AbstractGrant

andrewgodwin:
What difficulty level would we be looking at for stateless authorization codes? (and what sort of length if they were manually input).

cosmin:
Not hugely difficult, we'd could just encode the authorization object to a string plus an hmac. But I think that kind of design tends to be mostly useful for high volume services. 

Given the small-ish instance focus of this project I think the simple design of storing the authorization object is more appropriate. Is the extra table a concern?

andrewgodwin:
I'm just mostly interested in avoiding extra tables wherever possible because of personal preference (and also one less table to have to go do cleanup jobs on). If you think it's feasible to do it all in the auth string and keep it relatively short (under 64 chars, say) then let's do that.

cosmin:
It wouldn't relatively short, the relatively short option is to store the data in the database. Embedding all the data in the authorization code instead is the relatively long option, the HMAC alone would be 44 characters when base64 encoded.

andrewgodwin:
Ah, fine, I'll capitulate and do the sensible thing of having it in a table :)

andrewgodwin (approved):

PR #342, opened 2023-01-01
Don’t run ‘publish dev image’ job in forks

aviflax:
My fork (and, I assume, most forks) doesn’t have the secrets defined, so this job just fails every night. So it’ll be less noisy and less wasteful of resources to disable this job for forks.


andrewgodwin (approved):

andrewgodwin:
Yes, very sensible, thanks!

Issue #343, opened 2023-01-01
Uploading media with local media backend Breaks the pages that tries to display it


humrochagf:
Hello! I'm using `local` media backend both on development and in my takahe deployment.

After I upload a profile picture to my account, all pages that try to display it fail with this error:

```

Internal Server Error: /
Traceback (most recent call last):
  File "/home/username/Projects/takahe/.venv/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/home/username/Projects/takahe/.venv/lib/python3.11/site-packages/django/core/handlers/base.py", line 220, in _get_response
    response = response.render()
               ^^^^^^^^^^^^^^^^^
  File "/home/username/Projects/takahe/.venv/lib/python3.11/site-packages/django/template/response.py", line 114, in render
    self.content = self.rendered_content
                   ^^^^^^^^^^^^^^^^^^^^^
  File "/home/username/Projects/takahe/.venv/lib/python3.11/site-packages/django/template/response.py", line 92, in rendered_content
    return template.render(context, self._request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/username/Projects/takahe/.venv/lib/python3.11/site-packages/django/template/backends/django.py", line 62, in render
    return self.template.render(context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/username/Projects/takahe/.venv/lib/python3.11/site-packages/django/template/base.py", line 175, in render
    return self._render(context)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/username/Projects/takahe/.venv/lib/python3.11/site-packages/django/test/utils.py", line 111, in instrumented_test_render
    return self.nodelist.render(context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/username/Projects/takahe/.venv/lib/python3.11/site-packages/django/template/base.py", line 1005, in render
    return SafeString("".join([node.render_annotated(context) for node in self]))
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/username/Projects/takahe/.venv/lib/python3.11/site-packages/django/template/base.py", line 1005, in <listcomp>
    return SafeString("".join([node.render_annotated(context) for node in self]))
[...]

PR #344, opened 2023-01-01
Fix absolute url error while uploading media with local media backend

humrochagf:
When using local media backend, the image doesn't have the fqdn part of the URL, so I adapted the `AutoAbsoluteUrl` to be able to deal with both CDN based media and local media automatically.

Fixes: #343

### DIFF
--- a/activities/models/post_attachment.py
+++ b/activities/models/post_attachment.py
@@ -3,7 +3,7 @@ from functools import partial
 from django.db import models
 
 from core.uploads import upload_namer
-from core.uris import ProxyAbsoluteUrl, RelativeAbsoluteUrl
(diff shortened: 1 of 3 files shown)


andrewgodwin:
So this is actually a deliberate design decision - media URLs _need_ to have an absolute path, and I want this to be provided in the settings in case we use a route for media that does not go through these URL methods.

If you want to enact a similar automatic change, you could instead put in a PR for the settings file that adds MAIN_DOMAIN to the start of MEDIA_URL if it doesn't have a domain set?

humrochagf:
The change made to ask for the proper setting of media_url in https://github.com/jointakahe/takahe/commit/43ba09a16e68fd0633fa4a20ad5aa5dd2728bdcd is the best solution, better to ask for the good config than assuming the MAIN_DOMAIN :smiley:

PR #345, opened 2023-01-01
Apply image orientation metadata before resizing.

### LINKED ISSUE #341: Sometimes when I upload an image, it is turned upside-down
I’m not 100% sure but I think this has happened either exclusively, or mostly, when I upload an HEIF image in Mobile Safari via the “Photo Library” picker. I’ll try to start keeping track of when this happens and when it doesn’t.

Here’s a screenshot illustrating this:

54485BB1-F673-4559-A685-B742AF5D31C0

At the time I was running on da00a67, and using the Dockerfile so I’m pretty sure this isn’t a Python dependency management problem. (Or at least, what I mean is, at least it’s not a case of me installing a different set of deps than the code intends, that kind of thing.)

TkTech:
Closes #341

### DIFF
--- a/core/files.py
+++ b/core/files.py
@@ -24,6 +24,16 @@ def resize_image(
     to fit if needed)
     """
     with Image.open(image) as img:
+        try:
+            # Take any orientation EXIF data, apply it, and strip the
+            # orientation data from the new image.
+            img = ImageOps.exif_transpose(img)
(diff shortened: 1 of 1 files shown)


andrewgodwin (approved):

andrewgodwin:
Ta!

PR #347, opened 2023-01-02
Add Domain.state with nodeinfo fetching

manfre:


users/models/domain.py
```diff
@@ -1,10 +1,45 @@
...
+    purged = State()
+
+    outdated.transitions_to(updated)
+    updated.transitions_to(outdated)
+
+    outdated.transitions_to(connection_issue)
+    outdated.transitions_to(purged)
+    connection_issue.transitions_to(purged)
+
+    outdated.times_out_to(connection_issue, 60 * 60 * 12)
```
andrewgodwin, line 27: We need a way back from connection_issue - maybe just give it a much longer `try_interval` (e.g. 24 hours) and then if something comes back we'll eventually realise?

Also let's make the timeout on going to this state 24 hours, and the try interval on outdated 30 mins.

manfre: I'll update the outdated intervals, but my thought process for the connection_issue state was to have it sit in that state until manual intervention (admin "force state"), at least until we have more tooling in place to better determine if it's an outage or the other server is unsupported.

andrewgodwin (approved):

PR #348, opened 2023-01-02
Include an /etc/mime.types file to correct gaps from mimetypes package

manfre:
Make it so `mimetypes.guess_type("somefile.webp")` returns `image/webp`


andrewgodwin (approved):

andrewgodwin:
Nice catch!

PR #349, opened 2023-01-04
Basic invitation tracking

dcwatson:
While I'd like to explore invitation trees and how they could be useful for moderation, and eventually allow users to create invitations, that all depends on some basic building blocks -- tracking who created an invitation, and whose invitation link you used when signing up. This PR is mostly just to get the migrations in early to start building the tree, even if there is no frontend yet.


andrewgodwin:
I'd like to avoid any further changes to the invite system just yet - this is a bit beyond the scope of what I had planned.

dcwatson:
No worries, happy to let this sit, but I wanted to clarify that I don't have any plans for this PR beyond just tracking creators/users of invitations. Regardless of UI, I think it's a useful question to be able to answer once your instance is of a size that you deputize moderators to create invitations.

andrewgodwin:
So I'd like to not have invited_by done like this ideally, and instead work on our UserEvent system that's supposed to be the audit log of all moderation- and server-level things that happen to a user. Given that's not ready yet, if you want to change this PR to just track the creator of invitations, I'd take that.

okpierre:
Thought of submitting a feature request until I saw this

For invite/referrals, it'll be nice to get a view of connections. 

If JohnDoe invites 3 legit users and 5 spammers, admin/mod can focus on those connections

Not sure if it's similar to the discussion here. Can open a separate issue if needed

andrewgodwin:
I'm not going to take this in for now, as I'm not really looking to build out a powerful invite system anyway - probably will just limit invites to admins rather than a social-graph thing.

PR #351, opened 2023-01-04
Re-use page for interaction query

tabletcorry:
Original code caused the query on timeline to issue twice. Once to satisfy the interactions lookup (which had no LIMIT) and then again for the page (which had a LIMIT 25). Presumably we want interactions for the paginated events, especially since the un-LIMITed query would become extremely inefficient.

With this fix, the query to get the timeline runs only once, with an appropriate LIMIT (and therefore the query to interactions is much smaller as well).

### DIFF
--- a/activities/views/timelines.py
+++ b/activities/views/timelines.py
@@ -25,14 +25,15 @@ class Home(TemplateView):
         events = TimelineService(self.request.identity).home()
         paginator = Paginator(events, 25)
         page_number = self.request.GET.get("page")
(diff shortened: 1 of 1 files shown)


tabletcorry:
I haven't (yet) looked to see if this pattern is present anywhere else, so it is something to keep an eye out for.

andrewgodwin (approved):

andrewgodwin:
Well that probably explains a lot about why that view was slow!

PR #352, opened 2023-01-04
Enable HTTPX integration in Sentry

tabletcorry:
I don't know a whole lot about HTTPX, but it seems Sentry has an integration for it. Should it be enabled?


tabletcorry:
Please feel free to close this if it doesn't really make any sense. Raised it in case it has some relevance :)

andrewgodwin (approved):

andrewgodwin:
Makes sense to me!

PR #355, opened 2023-01-05
Improve /api/v1/accounts/{id}/statuses perf

tabletcorry:
PR takes a few main paths:

(1) Use Django-native features such as `select_related` and `prefetch_related` to radically reduce query count

(2) Improve `Post.to_mastodon_json` to more carefully load only required data, and optimize for common case

(3) Introduce memoization on certain functions. This is critically useful on `Identity.to_mastodon_json` which is a very expensive function. It isn't clear that there is any other way to cheapen this function, so a basic memoization is extremely effective

Notes:
(a) If memoization is not desirable by the maintainers, it can be removed and this PR will still return substantial improvements

(b) If mild denormalization is acceptable, storing `reply_parent.author.pk` on the `Post` would yield a moderate query load reduction

(c) The followers/following counts on the `Identity` are a main issue for DB optimization there. Storing an async count on the `Identity` model would likely solve this, at the cost of adding more Stator work (and denormalization)


tabletcorry:
Also, the cache timing on the memoization methods is definitely placeholder. But wanted to gather feedback on memoization before I addressed that 😄

users/models/identity.py
```diff
@@ -812,6 +813,16 @@ def to_opengraph_dict(self) -> dict:
 
     ### Mastodon Client API ###
 
+    @cache_memoize(120)
+    def mention_to_mastodon_json(self):
```
andrewgodwin, line 817: This should be called `to_mastodon_mention_json`

andrewgodwin (commented):
I don't want the cache memoizer in here - there should be a more elegant way of solving this than that, via select_related and annotate. Is the following/followers counts the only repeated query left after this?

tabletcorry:
Further discussion in Discord: https://discord.com/channels/1044001375668936785/1044080967037497425/1060418889630109716

activities/models/emoji.py
```diff
@@ -258,6 +259,7 @@ def by_ap_tag(cls, domain: Domain, data: dict, create: bool = False):
 
     ### Mastodon API ###
 
+    @cache_memoize(120)
```
manfre, line 262: `cache_memoize` makes this method slower. We've already paid the cost of fetching the instance and `Config.system` is loaded at the start of requests and periodically by Stator.

users/models/identity.py
```diff
@@ -812,6 +813,16 @@ def to_opengraph_dict(self) -> dict:
...
+        return {
+            "id": self.id,
+            "username": self.username or "",
+            "url": self.absolute_profile_uri() or "",
+            "acct": self.handle or "",
+        }
+
+    @cache_memoize(120)
```
manfre, line 825: This is only used by Identity search from the clients. How likely are users on the same instance to search for the same Identity within a 2 minute window. I think this may be an incurred cost without any realized benefit.

manfre (changes requested):
I think the memoize here adds a few bugs and performance issues.

* `Post.to_mastodon_json` will have poor (and potentially unclear) cache behaviors due to the `interactions` param. The flow of create a new Post and then immediately view it will result in 2 separate cache keys (with and w/o interactions). Viewing a threaded discussion will result in each level/context having it's own cached representation of each Post.
* The lack of cache invalidation means the wrong record information will potentially be returned for an update/delete API calls.
* Many of the memoized `to_mastodon_json` methods are within loops. With a shared network cache (common case for larger sites that would benefit from caching), this means more IO than would have existed without the caching.

tabletcorry:
@manfre Agreed, and the memoize use I had was intended as a test of the waters. The discussion in Discord fleshed that out, and I will modify it accordingly.

Though as I discussed in Discord, prefetch may not solve all the issues here. Some of the queries are done independently and don't stack into the initial query set (though I will try to get it working).

users/models/identity.py
tabletcorry: Will be removed

users/models/identity.py
```diff
@@ -812,6 +813,16 @@ def to_opengraph_dict(self) -> dict:
 
     ### Mastodon Client API ###
 
+    @cache_memoize(120)
```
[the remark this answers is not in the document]
tabletcorry: Will be removed

api/views/accounts.py
```diff
@@ -133,8 +133,8 @@ def account_statuses(
     queryset = (
         identity.posts.not_hidden()
         .unlisted(include_replies=not exclude_replies)
-        .select_related("author")
-        .prefetch_related("attachments")
+        .select_related("author", "author__domain")
+        .prefetch_related("attachments", "mentions", "emojis")
```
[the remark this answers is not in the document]
tabletcorry: Nod, will try that.

activities/models/post.py
```diff
@@ -980,14 +987,7 @@ def to_mastodon_json(self, interactions=None):
...
             "mentions": [
-                {
-                    "id": mention.id,
-                    "username": mention.username or "",
-                    "url": mention.absolute_profile_uri() or "",
-                    "acct": mention.handle or "",
-                }
-                for mention in self.mentions.all()
-                if mention.username
+                mention.mention_to_mastodon_json() for mention in self.mentions.all()
```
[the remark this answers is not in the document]
tabletcorry: Good question. I wasn't entirely sure why there would be a mention without a username, and the inner code handles the null case independently.

users/models/identity.py
tabletcorry: Fixed

api/views/accounts.py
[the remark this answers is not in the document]
tabletcorry: Fixed

tabletcorry:
Ok! Updated based on feedback! Please see new summary on the PR for details.

There is still some room for improvement, either with annotations (which I don't yet understand) or some Stator denormalization (which is beyond the scope of this PR).

A concern I have is that:

```py
queryset = (
        identity.posts.not_hidden()
        .unlisted(include_replies=not exclude_replies)
        .select_related("author", "author__domain")
        .prefetch_related(
            "attachments",
            "mentions__domain",
            "emojis",
            "author__inbound_follows",
            "author__outbound_follows",
            "author__posts",
        )
        .order_by("-created")
```

Specifically `author__posts` will cause the query to load _all_ of the author's posts, which is plainly very bad. This is perhaps the strongest candidate for Stator denormalization (though follows would be a very close second).

activities/models/post.py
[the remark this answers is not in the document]
tabletcorry: @andrewgodwin Are you familiar with why the `if` was there? IIRC the `if` check being in the list comp caused an extra query, but I would need to check.

tabletcorry:
I have an idea for something to try, so please don't merge this yet (comment still welcome, of course).

tabletcorry:
Nevermind. I think it may make more sense to push the prefetches into the paginator, but that can be a followup PR.

activities/models/post.py
```diff
@@ -959,7 +959,12 @@ def to_opengraph_dict(self) -> dict:
     def to_mastodon_json(self, interactions=None):
         reply_parent = None
         if self.in_reply_to:
-            reply_parent = Post.objects.filter(object_uri=self.in_reply_to).first()
+            # Load the PK and author.id explicitly to prevent a SELECT on the entire author Identity
+            reply_parent = (
```
andrewgodwin, line 963: How much is this actually saving? It's not going to be a ton I think, and it's making the rest of the code harder to read (and if you just want to avoid loading all the columns, use `only` rather than `values`)

tabletcorry: Ah, good to know. I was looking for something like `only` and only found `values`

tabletcorry: Fixed!

andrewgodwin (approved):

andrewgodwin:
Thanks for the revisions on this! We'll attack the other perf issues separately - I think having a "lite=True" option to identity.to_mastodon_json that avoids follow counts might be a good start, and always use that when it's called from Post.

PR #358, opened 2023-01-05
Add Makefile targets for some common/useful development tasks:

### LINKED ISSUE #357: Add standard make targets for local python development
As a developer, it would be nice if the "local setup" were more streamlined, so that we don't have to keep switching back and forth between the [Contributing](https://docs.jointakahe.org/en/latest/contributing/) page and a terminal.

I typically (and would suggest, minimally) creating  `Makefile` targets for:

- `'venv` - create the virtual environment (includes installing dependencies though sometimes I break those out into their own targets too)
- `.git/hooks/pre-commit` - install the pre-commit hooks for development
- `test` - run the pytests

I also like to wrap these in useful general targets:

- Something like `pydev` - create the virtual environment, install pre-commit hooks, etc
- Something like `startdb` to get the DB up and running (not always, it's sometimes useful)
- Targets to run different services

sivy:
## Development Make Targets

This adds several new `Makefile` targets to help streamline local development.

One decision I made on this was to use the docker-compose postgres rather than assuming a locally-hosted Postgres. Might not be everyone's choice.

### dev setup tasks:

- `.venv`
- /git/hooks/pre-commit
- .env

### db tasks: reuses the docker db container rather than hosting postgres locally:

- startdb
- stopdb
- migrations

### active development tasks:

- server
- stator

## Usage/Example

```
$ make pydev
python3 -m venv .venv
. .venv/bin/activate
python3 -m pip install -r requirements-dev.txt
[snip install stuff...]
Successfully installed Jinja2-3.1.2 MarkupSafe-2.1.1 Pygments...
[snip install stuff...]
python3 -m pre_commit install
pre-commit installed at .git/hooks/pre-commit
```

```
make server
docker compose -f docker/docker-compose.yml up db -d
[+] Running 1/0
 ⠿ Container docker-db-1  Running                                                                                                        0.0s
python3 -m manage runserver
Watching for file changes with StatReloader
Performing system checks...

System check identified no issues (0 silenced).
January 05, 2023 - 16:33:20
Django version 4.1.5, using settings 'takahe.settings'
Starting development server at http://127.0.0.1:8000
```

Closes #357

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,15 +25,10 @@ compose_down:
 	python3 -m pre_commit install
 
 .env:
-	# afterwords, set in .env
-	# TAKAHE_DATABASE_SERVER="postgres://postgres:insecure_password@localhost:5433/takahe"
 	cp development.env .env
 
-_PHONY: pydev
-pydev: .venv
-
-_PHONY: precommit
-precommit: pydev .git/hooks/pre-commit
+_PHONY: setup_local
+setup_local: .venv .env .git/hooks/pre-commit
 
 _PHONY: startdb stopdb
 startdb:
@@ -43,23 +38,20 @@ stopdb:
 	docker compose -f docker/docker-compose.yml stop db
 
 _PHONY: superuser
-createsuperuser: .env pydev startdb
+createsuperuser: setup_local startdb
 	python3 -m manage createsuperuser
 
-_PHONY: pydev
-pydev: pydev precommit
-
 _PHONY: test
-test: pydev
+test: setup_local
 	python3 -m pytest
 
 # Active development
 _PHONY: migrations server stator
-migrations: startdb
+migrations: setup_local startdb
 	python3 -m manage migrate
 
-runserver: startdb
+runserver: setup_local startdb
 	python3 -m manage runserver
 
-runstator: startdb
+runstator: setup_local startdb
 	python3 -m manage runstator
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ stopdb:
 	docker compose -f docker/docker-compose.yml stop db
 
 _PHONY: superuser
-superuser: .env pydev startdb
+createsuperuser: .env pydev startdb
 	python3 -m manage createsuperuser
 
 _PHONY: pydev
@@ -58,8 +58,8 @@ _PHONY: migrations server stator
 migrations: startdb
 	python3 -m manage migrate
 
-server: startdb
+runserver: startdb
 	python3 -m manage runserver
 
-stator: startdb
+runstator: startdb
 	python3 -m manage runstator
--- a/Makefile
+++ b/Makefile
@@ -29,11 +29,11 @@ compose_down:
 	# TAKAHE_DATABASE_SERVER="postgres://postgres:insecure_password@localhost:5433/takahe"
 	cp development.env .env
 
-_PHONY: python-dev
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -14,3 +14,52 @@ compose_up:
...
+
+_PHONY: startdb stopdb
+startdb:
+	docker compose -f docker/docker-compose.yml up db -d
+
+stopdb:
+	docker compose -f docker/docker-compose.yml stop db
+
+_PHONY: superuser
+superuser: .env python-dev startdb
```
andrewgodwin, line 46: Let's call this createsuperuser

Makefile
```diff
@@ -14,3 +14,52 @@ compose_up:
...
+.env:
+	# afterwords, set in .env
+	# TAKAHE_DATABASE_SERVER="postgres://postgres:insecure_password@localhost:5433/takahe"
+	cp development.env .env
+
+_PHONY: python-dev
+python-dev: .venv
+
+_PHONY: precommit
+precommit: python-dev .git/hooks/pre-commit
```
andrewgodwin, line 36: Why is this `python-dev`?

Makefile
```diff
@@ -14,3 +14,52 @@ compose_up:
...
+_PHONY: test
+test: pydev
+	python3 -m pytest
+
+# Active development
+_PHONY: migrations server stator
+migrations: startdb
+	python3 -m manage migrate
+
+server: startdb
```
andrewgodwin, line 61: Can these be `runserver` and `runstator` instead? I like makefile actions to sound more like what they're doing.

Makefile
```diff
@@ -14,3 +14,52 @@ compose_up:
...
+# Development Setup
+.venv:
+	python3 -m venv .venv
+	. .venv/bin/activate
+	python3 -m pip install -r requirements-dev.txt
+
+.git/hooks/pre-commit: .venv
+	python3 -m pre_commit install
+
+.env:
```
andrewgodwin, line 27: Can we combine this into the venv one and call it `setup_local` or something? It's a bit too granular right now.

Makefile
sivy: Sorry switched from `python-dev` to `pydev` halfway through hashtag-adhd

Makefile
sivy: sure thing

Makefile
sivy: Done

Makefile
sivy: I wouldn't mind something like:

```
_PHONY: setup_local
setup_local: .venv .env .git/hooks/pre-commit
```

And remove the current `precommit` target, then make everything else (that needs to) depend on `setup_local`. I do like to keep targets for specific files when there are actual files to be built.

Thoughts?

sivy: I've made these changes, as well as switching the `development.env` template file to use the docker-hosted postgres by default.

andrewgodwin (approved):

andrewgodwin:
Thanks for these - do you want to take a stab at writing the documentation for this as well?

sivy:
Sure, in `contributing.md`?

andrewgodwin:
No, in `docs/contributing.rst` if you could - that's where our detailed docs for local dev live.

sivy:
Yeah that's what I meant, was writing from (bad) memory.

PR #359, opened 2023-01-06
Update emoji for domain when shortcode doesn't match uri

manfre:
This handles the situation where the remote replaces an emoji shortcode with a new record/object_url.

### DIFF
--- a/stator/models.py
+++ b/stator/models.py
@@ -253,6 +253,25 @@ class StatorModel(models.Model):
 
     atransition_perform = sync_to_async(transition_perform)
 
+    def transition_set_state(self, state: State | str):
+        """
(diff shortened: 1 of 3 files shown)


stator/models.py
```diff
@@ -253,6 +253,25 @@ def transition_perform(self, state: State | str):
 
     atransition_perform = sync_to_async(transition_perform)
 
+    def set_transition_state(self, state: State | str):
```
andrewgodwin, line 256: Can you rename this to `transition_set_state` so the name is consistent with a leading `transition`?

It would be nice not to duplicate this logic, but I see why you've done it (to save the extra UPDATE).

PR #362, opened 2023-01-06
Initial pass at hyperscript modal image popup

manfre:
This provides basic modal image viewing. Draft PR because the CSS is very rough right now.

I have this PR running on https://manfre.net and the logged out view is not as bad as the signed in timeline.


manfre:
Closing in favor of #363

PR #368, opened 2023-01-07
Adjust icon position on the right sidebar

shuuji3:
Right icons were not centered slightly on the small screen. This is because each icon has a bit different width and they were aligned on the left side. (There is no issue with the large screen.)

takahe-adjust-icon-position


andrewgodwin:
Thanks for fixing this, I noticed it during the week but didn't get to it yet!

PR #370, opened 2023-01-07
Fix left column height on small screen

shuuji3:
This change fixes where there is no margin at the bottom of the left columns. The cause was the current CSS set a small height for the left column when the content is more than one viewport.

**(before -> after)**
takahe social_compose_(iPhone SE) localhost_8000_compose_(iPhone SE)

### DIFF
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -1549,11 +1549,6 @@ form .post {
 
 @media (max-width: 920px),
 (display-mode: standalone) {
-    div.columns {
-        height: calc(100vh - 50px);
-        align-items: stretch;
-    }
-
     .left-column {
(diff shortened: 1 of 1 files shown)


static/css/style.css
```diff
@@ -1549,11 +1549,6 @@ form .post {
 
 @media (max-width: 920px),
 (display-mode: standalone) {
-    div.columns {
-        height: calc(100vh - 50px);
-        align-items: stretch;
```
shuuji3, line 1554: `align-items: stretch;` can be deleted safely because the same one is defined here: https://github.com/jointakahe/takahe/pull/370/files#diff-a5ea33c888430601a659bcbef2da1944097953737f2606282efc13ca6e5fbabaL338

andrewgodwin:
Thanks!

PR #373, opened 2023-01-08
Allow better keyboard navigation

### LINKED ISSUE #156: [Accessibility] Cannot focus several buttons with Tab key
When I try to navigate the site only using a keyboard, I noticed that several buttons are skipped with the Tab key. I'm using the latest Google Chrome on macOS.

## "Add image" button
Screenshot 2022-12-10 at 19 50 48

## "CW" button
Screenshot 2022-12-10 at 19 54 20

## menu items
Screenshot 2022-12-12 at 1 53 31

I guess the following are the causes:
- "Add image" button is `<a>` tag with `button` class
- "CW" button is `<span>` tag with `button` class
- All menu items are `<a>` tag but has no `href` attribute except for the "Reply" menu

Probably we can change the first two elements to `<button>` to allow focusing, but I'm not sure the last case. Maybe some accessibility guide line explains how to handle a case like this.

shuuji3:
fix #156

This change allows all items to be focusable and to interact with the keyboard. I tested whole pages and there should be no unfocusable item by keyboard.

Here are some findings to be accessible by keyboard:
- The browser disables focus on the `<a>` element without `href` so we need to add `tabindex="0"` to be focusable (ref. https://www.a11yproject.com/posts/how-to-use-the-tabindex-attribute/)
- hyperscript and htmx's click event is not triggered by the Enter key so we need to add an additional event like `keyup[key is 'Enter']` in addition to `on click`.

### DIFF
--- a/templates/activities/post_delete.html
+++ b/templates/activities/post_delete.html
@@ -7,7 +7,7 @@
     {% include "activities/_mini_post.html" %}
     <form action="." method="POST">
         {% csrf_token %}
-        <a class="button" onclick="history.back()">Cancel</a>
+        <button onclick="history.back()" autofocus>Cancel</button>
         <button class="delete">Delete</button>
     </form>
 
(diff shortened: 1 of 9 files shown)


templates/activities/post_delete.html
```diff
@@ -7,7 +7,7 @@ <h1>Delete this post?</h1>
     {% include "activities/_mini_post.html" %}
     <form action="." method="POST">
         {% csrf_token %}
-        <a class="button" onclick="history.back()">Cancel</a>
+        <button onclick="history.back()" autofocus>Cancel</button>
```
shuuji3, line 10: This `autofocus` is just for usability improvement similar to common dialog.

Screenshot 2023-01-09 at 2 06 26

andrewgodwin (approved):

andrewgodwin:
Thanks!

PR #384, opened 2023-01-09
Add side margin to home and login page on small screen

### LINKED ISSUE #369: Login page has no side margin on small screen
It would be nice to put a small margin next to "LOGIN" header and around both buttons the same as the form input.

Screenshot 2023-01-08 at 0 19 18

shuuji3:
resolves #369

(before / after)
 

This also fixes another same issue on the home page.

(before / after)

### DIFF
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -1607,8 +1607,11 @@ form .post {
         font-size: 22px;
     }
 
-    .left-column {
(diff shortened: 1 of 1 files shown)


andrewgodwin (approved):

andrewgodwin:
Thanks!

PR #386, opened 2023-01-09
Add Azure Blob Storage as a media option

tebriel:
This is a Draft PR while I test out the configuration and button up settings.

I'm not sure if this is something you'd want to include, but I'm using Azure to host all my infrastructure and I've previously used django-storages[azure] to host media in Azure Blob Storage, I believe this will work without much extra effort (haha famous last words). More updates and "ready for review" will come after some testing.


andrewgodwin:
Yup, I'd like to have this support! Just leave a comment when you're ready for review.

rcarmo:
I was looking into this as well, so allow me to chime in regarding two possible modes of operation:

1. You store the meda inside a storage account container and proxy the requests through the Takahe front-end
2. You can store the media inside the `$web` container and just redirect (or rewrite URLs) to have the browser fetch them directly from storage (and you can have that mapped to your own domain). This avoids the need for public blob access policies, since `$web` is public by default, and if you are using a domain-centric CDN, you can have that cover both your container and your public site.

I had a look at the [azure storage provider](https://github.com/jschneier/django-storages/blob/master/storages/backends/azure_storage.py) and it _seems_ to only support both modes of operation, but I have never actually used it - I just thought it would be a good idea to point out the differences and explain that it would be more "natural" to use the `$web` container and allow for a custom domain to be mapped (like `media.instance.domain` instead of the default Azure ones).

tebriel:
@rcarmo thanks! I got this working in https://github.com/bookwyrm-social/bookwyrm and am trying to apply that here but I've gotten stymied a few times because I've just switched to an m1 mac and the loss of arm64 docker builds locally is slowing my work down (though I should really just set up a true local dev env). I think it's fine to set the base URL for the assets to be the azure blob URL, at least as a first run, the django storage provider for azure also supports custom URL settings, so those will hopefully be easy to set as well (though my blob storage container is just bare azure blob url so I'll have to set that up to test it!)

rcarmo:
I gave up on Docker Desktop a long while ago... I just use Parallels with an Ubuntu ARM VM and either set DOCKER_HOST on the local CLI (which you can get as a standalone binary) or VSCode remote to it -- it is _way_ faster, although you do need to get used to your working source tree being "remote" (which has never been a real issue for me).

Anyway, I will be migrating my instance to Azure soon (famous last words), will eventually have time to test this.

tebriel:
Awesome, I'm using Colima locally and it's fine for arm but it's an M1 Air so not much power. I got my local non-docker dev setup yesterday before I went out to play some board games. Can now repro my auth issue locally and just need to tweak it to figure out why it's not using the creds correctly!

.github/workflows/docker-dev.yml
```diff
@@ -30,9 +30,9 @@ jobs:
         id: meta
         uses: docker/metadata-action@57396166ad8aefe6098280995947635806a0e6ea
         with:
-          images: jointakahe/takahe-dev
+          images: cmoultrie/takahe-dev
```
AstraLuma, line 33: This entire file should not have been committed.

AstraLuma (changes requested):
The meat of this visually looks good, but you seem to have left some personal environment detritus in the PR.

tebriel:
I no longer run takahe.

PR #387, opened 2023-01-10
Improve performance on /api/v1/timelines/home

tabletcorry:
Prefetch/select required fields for API version of home. Requires considerably more data than HTML version of home.

Results:
* Before 699.486 [ms] (>100+ SQL queries)
* After 376.082 [ms] (~16 SQL queries)

### DIFF
--- a/api/views/timelines.py
+++ b/api/views/timelines.py
@@ -21,6 +21,30 @@ def home(
     # Grab a paginated result set of instances
     paginator = MastodonPaginator()
     queryset = TimelineService(request.identity).home()
+    queryset = queryset.select_related(
+        "subject_post_interaction__post",
+        "subject_post_interaction__post__author",
+        "subject_post_interaction__post__author__domain",
+    )
+    queryset = queryset.prefetch_related(
+        "subject_post__mentions__domain",
+        "subject_post__author__inbound_follows",
+        "subject_post__author__outbound_follows",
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Are you sure the `author__inbound_follows` and similar prefetches are still needed? I updated the JSON methods for Post/PostInteraction to (I thought) not return follower counts over the weekend.

tabletcorry:
@andrewgodwin Not completely sure, I was pretty liberal with the fetches, and copied out of the last API method I did.

andrewgodwin:
Yeah, the fetches do have their own overhead which is why I'm a bit hesistant. Could you remove the `__posts`, `__inbound_follows` and `__outbound_follows` additions for now, and I'll poke at it a bit on my end later?

tabletcorry:
> Yeah, the fetches do have their own overhead which is why I'm a bit hesistant. Could you remove the `__posts`, `__inbound_follows` and `__outbound_follows` additions for now, and I'll poke at it a bit on my end later?

Removed and tested. Looks like they were no longer required 👍

tabletcorry:
Oh, you mentioned posts too. Let me try that

andrewgodwin:
Lovely. That test failure looks unrelated so I'll merge this in and fix any test issues on main myself!

PR #389, opened 2023-01-11
Introduce index for timelineevents published

tabletcorry:
Rendering the home timeline is a fairly intense query, and is used in several of the heaviest APIs in the system. Currently they are sorted by -published which is not indexed, so the query planner tries various ways to get the data as best as possible.

With an explicit index though, it can readily grab the most recent events, and move on.

Example stats from my takahe.tabletcorry.com server on my user, rendering the home timeline:

Before:
     Planning Time: 6.120 ms
     Execution Time: 12.109 ms
After:
    Planning Time: 4.797 ms
    Execution Time: 0.769 ms

This improvement is not super visible in benchmarking due to other noise dominating, but it will doubtless reduce load on the DB overall.


andrewgodwin:
Apologies for the slow roll on this one, but I think I have decided to follow Mastodon's path here and show posts by created (received) order, rather than by published order, so we won't need this PR!

PR #397, opened 2023-01-11
Fix identity not fetching emoji

humrochagf:


humrochagf:
We already do like this for the posts, thats why, some of the emoji creation happens when receiving a post, but not from identity https://github.com/jointakahe/takahe/blob/main/activities/models/post.py#L795

andrewgodwin (approved):

andrewgodwin:
Aha! That's why it used to work and now doesn't! We changed the namespacing :)

PR #407, opened 2023-01-13
Admin action to copy Emoji to local

manfre:
Also deletes the underlying file when deleting an Emoji in the Admin.


andrewgodwin:
Could you move `copy_to_local` to be on the Emoji model? I'd like to call it from the new emoji admin pages as well.

manfre:
@andrewgodwin I moved it `Emoji` and implemented copy for the new emoji admin.

andrewgodwin (approved):

PR #382, opened 2023-01-09
Add djhtml to pre-commit check

shuuji3:
I noticed several HTML files have inconsistent indentation (like 2 or 4 spaces).

This PR adds djhtml, which can understand Django template syntax and ensure 4 spaces indentation.

djhtml repository: [rtts/djhtml: Django/Jinja template indenter](https://github.com/rtts/djhtml)

I also tried another linter called djlint. This can inspect HTML and templates a bit more comprehensively but it's too aggressive and return too many issues:

```shell
djLint for Django........................................................Failed
- hook id: djlint-django
- exit code: 1


100% 11/11 [00:00<?, ?it/s]        


templates/about.html
───────────────────────────────────────────────────────────────────────────────
T003 2:24 Endblock should have name. Ex: {% endblock body %}. {% endblock %}
H006 5:8 Img tag should have height and width attributes. <img class="banner" 
H013 5:8 Img tag should have an alt attribute. <img class="banner" 
H017 5:8 Tag should be self closing. <img class="banner" 
T003 16:0 Endblock should have name. Ex: {% endblock body %}. {% endblock %}

templates/activities/follows.html
───────────────────────────────────────────────────────────────────────────────
T003 3:27 Endblock should have name. Ex: {% endblock body %}. {% endblock %}
H029 24:24 Consider using lowercase form method values. <form action="{{ ide
H017 27:32 Tag should be self closing. <input type="hidden"
H017 30:32 Tag should be self closing. <input type="hidden"
T003 60:0 Endblock should have name. Ex: {% endblock body %}. {% endblock %}

templates/admin/federation.html
───────────────────────────────────────────────────────────────────────────────
T003 2:30 Endblock should have name. Ex: {% endblock body %}. {% endblock %}
H017 5:8 Tag should be self closing. <input type="search"
T003 36:0 Endblock should have name. Ex: {% endblock body %}. {% endblock %}

templates/admin/federation_edit.html
───────────────────────────────────────────────────────────────────────────────
T003 2:39 Endblock should have name. Ex: {% endblock body %}. {% endblock %}
H029 4:4 Consider using lowercase form method values. <form action="." met
T003 19:0 Endblock should have name. Ex: {% endblock body %}. {% endblock %}

templates/admin/report_view.html
───────────────────────────────────────────────────────────────────────────────
T003 2:42 Endblock should have name. Ex: {% endblock body %}. {% endblock %}
H029 4:4 Consider using lowercase form method values. <form action="." met
[...]

PR #409, opened 2023-01-14
FIx HTML issues found by djLint

shuuji3:
This fixes various HTML issues found by djLint experiment in #382.

Each commits fixes the following types of issues:
- lack of spaces in the template variable
- single-quote characters for HTML attribute
- not using `{% url "..." %}` patterns
- unmatched HTML tags 
- usage obsoleted `<tt>` tag (ref. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tt)


andrewgodwin (approved):

andrewgodwin:
Lovely stuff.

PR #410, opened 2023-01-14
Fix emoji file size validation and help text and add image size validation

shuuji3:
Thanks for implementing the emoji admin page! 🙂 

I tried adding new emojis but it always returned "Image filesize is too large" error.

This PR fixes so that the size validation is performed correctly in KB, and amends the help text to show the actual file size limit in the settings.

Also, I added an image size (width/height, not file size) check as written in the help text.

## Validation error example
Screenshot 2023-01-14 at 20 44 37
Screenshot 2023-01-14 at 20 43 52

### DIFF
--- a/users/views/admin/emoji.py
+++ b/users/views/admin/emoji.py
@@ -49,13 +49,15 @@ class EmojiCreate(FormView):
             help_text="What users type to use the emoji :likethis:",
         )
         image = forms.ImageField(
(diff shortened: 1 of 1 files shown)


andrewgodwin (approved):

andrewgodwin:
Nice, sorry about the size validation mistake there!

shuuji3:
No problem! 200 bytes image was difficult to create 😃

PR #413, opened 2023-01-14
Instruct robots to ignore proxy

tabletcorry:
Robots are already crawling my takahē instance, which is fine, but crawling the proxy is a good way to burn bandwidth. Also Mastodon appears to have a similar block by default (e.g. https://mastodon.social/robots.txt)


andrewgodwin (approved):

andrewgodwin:
Good thinking!

PR #416, opened 2023-01-15
Prevent uploading too large image file

### LINKED ISSUE #392: Client-side image size check
To prevent Bad Gateway or other HTTP errors throwing away profile form changes, we should (if possible) have a client-side check that the file about to be uploaded is small enough too, as a backup.

shuuji3:
resolves #392

Fortunately, `input[type=file]` holds the size information of the selected image, which can be checked with hyperscript.

> `size`
> The size of the file in bytes as a read-only 64-bit integer.

ref. [Using files from web applications - Web APIs | MDN](https://developer.mozilla.org/en-US/docs/Web/API/File_API/Using_files_from_web_applications#getting_information_about_selected_files)

This PR will put the same error message DOM in the same place as Django generates. It also prevents clicking the "Upload" button in a similar way as the character count feature.

**Accepted case (<= 10MB)**
Screenshot 2023-01-15 at 14 30 53

**Rejected case (> 10MB)**
Screenshot 2023-01-15 at 14 30 33

(note: the orange color is green by default)

### DIFF
--- a/activities/views/compose.py
+++ b/activities/views/compose.py
@@ -204,7 +204,30 @@ class ImageUpload(FormView):
     template_name = "activities/_image_upload.html"
 
     class form_class(forms.Form):
-        image = forms.ImageField()
+        image = forms.ImageField(
+            widget=forms.FileInput(
+                attrs={
+                    "_": f"""
+                        on change
+                            if me.files[0].size > {settings.SETUP.MEDIA_MAX_IMAGE_FILESIZE_MB * 1024 ** 2}
+                                add [@disabled=] to #upload
+
+                                remove <ul.errorlist/>
+                                make <ul.errorlist/> called errorlist
+                                make <li/> called error
+                                set size_in_mb to (me.files[0].size / 1024 / 1024).toFixed(2)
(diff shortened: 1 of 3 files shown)


activities/views/compose.py
```diff
@@ -204,7 +204,30 @@ class ImageUpload(FormView):
...
+                    "_": f"""
+                        on change
+                            if me.files[0].size > {settings.SETUP.MEDIA_MAX_IMAGE_FILESIZE_MB * 1024 ** 2}
+                                add [@disabled=] to #upload
+
+                                remove <ul.errorlist/>
+                                make <ul.errorlist/> called errorlist
+                                make <li/> called error
+                                set size_in_mb to (me.files[0].size / 1024 / 1024).toFixed(2)
+                                put 'File must be 10MB or less (actual: ' + size_in_mb + 'MB)' into error
```
manfre, line 219: Hardcoded `10MB` should use `settings.SETUP.MEDIA_MAX_IMAGE_FILESIZE_MB`

andrewgodwin:
Lovely, thank you, I'll go update the hardcode myself!

PR #425, opened 2023-01-16
Add a css variable for Emoji height for easier customization

manfre:


andrewgodwin (approved):

andrewgodwin:
Hmm, shouldn't this actually be in `em` or `%` so that it's relative to text size?

andrewgodwin:
Cool, I'll merge it in and play with a relative size to see if it works.

PR #428, opened 2023-01-16
Admin improvements for Post and PostAttachment

manfre:


manfre:
Rebased to main and added a PostAttachment action to bulk guess/fix mimetypes.

PR #438, opened 2023-01-18
Show image at normal size when less than width 100%

manfre:
This `width: 100%;` causes images smaller than `width: 100%;` with a height < 400px to scale up to be 400px tall. This causes normally viewable images to either become blurry or clipped, forcing a click to view at normal scaling.

### DIFF
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -1512,7 +1512,6 @@ form .post {
     object-fit: cover;
     object-position: center;
     max-height: 400px;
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Maybe this should be switched to `max-width: 100%` instead?

manfre:
It already has `max-width: 100%`.

andrewgodwin:
So it does!

Issue #401, opened 2023-01-12
improve modularity in nginx config


foozmeat:
I'd like to get some of my boilerplate nginx config into the container but the current setup make that difficult. I wonder if Tekahe's config could be created as a set of overrides in `/etc/nginx/conf.d/` so I can manage my own overrides in the same way?

andrewgodwin:
I think that's feasible - it won't be high on my priority list, but I'll happily take a nice PR for it if it's not too many extra files (I'd like it to still be easy to read all together).

tabletcorry:
Would definitely be nice to have more control there. I have a PR that floats on top of main to add my own custom logging format.

PR #442, opened 2023-01-19
Nginx refactor for #401

foozmeat:
This PR does two things

1. let's someone drop their own vhost config into the container at /etc/nginx/conf.d/default.conf.tpl. Also, nginx is run without a config file path which is slightly clearer imo.
2. upgrades the container from buster to bullseye


foozmeat:
This is for #401

andrewgodwin (approved):

andrewgodwin:
Looks like a very reasonable set of changes - I'll test it out later!

PR #443, opened 2023-01-19
Downgrade most fetch related errors to capture_message

manfre:
I feel confident that we've hit the point where any HTTP level errors are not something we need to act upon and are safe to disable when not explicitly wanting to see them. I have this soaking on my instance to make sure it clears all the current nuisance errors.

### DIFF
--- a/activities/models/post.py
+++ b/activities/models/post.py
@@ -3,6 +3,7 @@ import hashlib
 import json
 import mimetypes
 import re
+import ssl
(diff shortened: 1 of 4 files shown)


andrewgodwin:
Agreed - it would also be nice to see if we can somehow group all JSON parsing/reading failures too so they all fall under one entry in Sentry, but that's probably something I'll look at over the weekend

PR #449, opened 2023-01-19
Organize nginx config files

tabletcorry:
(a) Remove gitignore for rendered files (they aren't a problem anymore)
(b) Drop error logging level to WARN because INFO is noisy without much benefit?
(c) Move default.conf.tpl to folder so additional files can be added in the future, or for development
(d) Add gitignore to folder to allow for development files to be kept locally

Of the above (b) is probably the only possibly controversial one, and I can drop it out if people care. But it seems to just spam noise?


andrewgodwin (approved):

andrewgodwin:
All seems sensible to me!

PR #453, opened 2023-01-19
Attach user identification headers to response

tabletcorry:
Add `X-Takahe-User` and `X-Takahe-Identity` headers to response, when available, to allow for better Nginx log enrichment.

Also drop these headers in Nginx so they aren't sent into the world. They probably aren't dangerous since they identfy the users _to themselves_ but strip it for now, just in case.

### DIFF
--- a/docker/nginx.conf.d/default.conf.tpl
+++ b/docker/nginx.conf.d/default.conf.tpl
@@ -25,6 +25,10 @@ server {
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     proxy_http_version 1.1;
 
+    # The user header is available for logging, but not returned to the client
+    proxy_hide_header X-Takahe-User;
+    proxy_hide_header X-Takahe-User-Identity;
+
     # Serve robots.txt from the non-collected dir as a special case.
     location /robots.txt {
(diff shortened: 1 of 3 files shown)


docker/nginx.conf.d/default.conf.tpl
```diff
@@ -25,6 +25,10 @@ server {
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     proxy_http_version 1.1;
 
+    # The user header is available for logging, but not returned to the client
+    proxy_hide_header X-Takahe-User;
+    proxy_hide_header X-Takahe-User-Identity;
```
manfre, line 30: It'll be easier to visually scan logs if you change it to `X-Takahe-Identity`

tabletcorry: Sure, changed. In my nginx logs, I don't log the actual header name so that didn't matter as much in my case, but sounds reasonable.

andrewgodwin:
Could you bundle this into the existing `IdentityMiddleware`? Each middleware we add involves a small performance hit.

andrewgodwin (approved):

andrewgodwin:
Thanks!

PR #459, opened 2023-01-20
Make menu icons visible on the profile page in the light theme

### LINKED ISSUE #446: Difficult to see header icons in light mode
The combination of dark color icons and dark background makes it difficult to see.

## Current
Screenshot 2023-01-19 at 18 42 38

### Option 1
Screenshot 2023-01-19 at 18 45 05

### Option 2
Screenshot 2023-01-19 at 18 43 03

shuuji3:
resolves #446

### DIFF
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -255,6 +255,10 @@ header menu a i {
     vertical-align: middle;
 }
 
+body.light-theme header menu a i {
+    color: var(--color-bg-main);
+}
+
 header menu a.htmx-request i {
     -webkit-animation-delay: var(--fa-animation-delay, 0s);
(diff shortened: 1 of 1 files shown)


shuuji3:
It was a bit tricky than imagined. The current change will fix the color on the profile page but it cannot be on other pages like home.

Profile page:
Screenshot 2023-01-21 at 1 26 29

Home page (icon color should be the same as dark mode):
Screenshot 2023-01-21 at 1 26 52

 It should be applied to the profile page but I'm not sure how we can distinguish the page from CSS. 🤔

shuuji3:
Thanks, that's what I needed to find. It seems working now.

andrewgodwin:
Thanks!

PR #465, opened 2023-01-20
Add favourites api endpoint

humrochagf:
Hey, this adds the endpoint in the API to fetch favourites.

I'm still trying to get the workflow of takahe's model/service/view so if I misplaced something just let me know :slightly_smiling_face:

### DIFF
--- a/activities/services/timeline.py
+++ b/activities/services/timeline.py
@@ -1,6 +1,6 @@
 from django.db import models
 
-from activities.models import Hashtag, Post, TimelineEvent
+from activities.models import Hashtag, Post, PostInteraction, TimelineEvent
 from activities.services import PostService
 from users.models import Identity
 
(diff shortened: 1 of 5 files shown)


humrochagf:
Another thing was that I named it as `favourites` to match the expected mastodon API, but we can rename internal functions the way it works best

humrochagf:
Also tested it with both Megalodon and Tusky apps :slightly_smiling_face: 
megalodon_favs
tusky_favs

api/views/timelines.py
```diff
@@ -144,3 +144,36 @@ def conversations(
...
+        limit=limit,
+    )
+    # Convert those to the JSON form
+    pager.jsonify_posts(identity=request.identity)
+
+    # Add the link header if needed
+    if pager.results:
+        response.headers["Link"] = pager.link_header(
+            request,
+            ["limit", "local", "remote", "only_media"],
```
andrewgodwin, line 176: This should just be "limit", as that's the only option on the endpoint

activities/services/timeline.py
```diff
@@ -82,3 +82,16 @@ def identity_public(self, identity: Identity):
             .unlisted(include_replies=True)
             .order_by("-id")
         )
+
+    def favourites(self) -> models.QuerySet[Post]:
```
andrewgodwin, line 86: Our convention is to call these `likes` everywhere except where we interface with Mastodon - so here and the tests should use like, and the mastodon API is the thing that's called favourites.

andrewgodwin (commented):
Looks pretty good, couple of small things!

api/views/timelines.py
humrochagf: fixed

activities/services/timeline.py
humrochagf: Alright, all updated :slightly_smiling_face:

andrewgodwin (approved):

andrewgodwin:
Thanks for this!

PR #474, opened 2023-01-26
Make user handle search to be case insensitive

humrochagf:
Mastodon API is case-insensitive when looking up user handles.


humrochagf:
With this change, it's easier to find users that have uppercase letters in their handle.

Before:
search-before

After:
search-after

andrewgodwin:
👍

PR #478, opened 2023-01-29
Templatize robots.txt

tabletcorry:
The existing static robots.txt did not allow any method of adding additional disallowed agents.

Since a substantial majority of web traffic to my test instance is a few SEO crawlers, and this is likely to be true for others as well, add a method by which well-behaving robots can be blocked.

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -2,6 +2,7 @@ import json
 from typing import ClassVar
 
 import markdown_it
+from django.conf import settings
 from django.http import HttpResponse
 from django.shortcuts import redirect
 from django.templatetags.static import static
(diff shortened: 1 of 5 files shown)


tabletcorry:
I don't think the pre-commit error is due to my change 🤔

shuuji3:
It looks like the pre-commit error is due to isort: https://github.com/PyCQA/isort/issues/2077

It was already fixed in v5.12.0, so I'll create a PR to upgrade the version.

andrewgodwin (approved):

andrewgodwin:
I was thinking about doing this anyway, so thanks!

tabletcorry:
The two that were annoying me the most for constant crawling are `TAKAHE_ROBOTS_TXT_DISALLOWED_USER_AGENTS='["AhrefsBot","SemrushBot"]'` in case anyone wants to copy a few easy ones.

PR #480, opened 2023-01-29
Upgrade isort version in pre-commit

shuuji3:
ref. #478

This newer version of isort should not have the error related to the py-shims version check.


andrewgodwin:
Thanks!

PR #483, opened 2023-01-29
Fix header duplication during infinity scroll

humrochagf:
The infinity scroll at the Identity page was duplicating the header in each pagination event

### DIFF
--- a/templates/identity/view.html
+++ b/templates/identity/view.html
@@ -15,7 +15,7 @@
 {% block body_class %}has-banner{% endblock %}
 
 {% block content %}
-    {% if not headless %}
+    {% if not request.htmx %}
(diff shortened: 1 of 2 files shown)


humrochagf:
Before:
Screenshot from 2023-01-29 18-55-44

After:
Screenshot from 2023-01-29 18-54-45

andrewgodwin (commented):
You can use `request.htmx` rather than adding `headless`!

humrochagf:
A nice feature from `django-htmx` I didn't know about it before :stuck_out_tongue_winking_eye:

andrewgodwin (approved):
Yeah, it's a really nice little package!

PR #472, opened 2023-01-24
Fix #471

okpierre:
-hide following and followers count
-exclude post count

https://github.com/jointakahe/takahe/issues/471

PR #484, opened 2023-01-30
Update test_privacy.py

okpierre:
Update test_privacy.py - target follower instead of div class

Thanks @shuuji3 https://github.com/jointakahe/takahe/pull/472#issuecomment-1403474833


shuuji3:
@okpierre Ah, you can add the above "Update test_privacy.py" commit to your `patch-2` branch, instead of this separate `patch-3` branch. Then, it would resolve the previous test failure in #472. 🙂

okpierre:
@shuuji3 I've made the change. Hope it is better now. Will close this. Thank you for your assistance

shuuji3:
You're welcome. 🙂

Issue #461, opened 2023-01-20
Add upgrade instructions in documentation


okpierre:
A quick search using 'update' and 'upgrade' in the documentation located here
`https://docs.jointakahe.org/en/latest/installation/`

yields only notes about release updates

For beginners and those new to Takahe, it would help to have instructions on how to upgrade from one version to the next

For example: upgrade from v6.x to v7.x or v7.x to the next

Maybe something like Mastodon?
`https://docs.joinmastodon.org/admin/upgrading/`

andrewgodwin:
Yes, it would be good to have a section for this, though it will just be:

- Bump the docker image version
- Run migrations

PR #486, opened 2023-01-30
Add documentation about upgrading Takahe

humrochagf:
Closes: #461

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -14,8 +14,9 @@ in alpha. For more information about Takahē, see
    :maxdepth: 2
    :caption: Contents:
 
-   installation
    features
+   installation
+   upgrading
    contributing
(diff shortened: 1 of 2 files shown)


docs/index.rst
```diff
@@ -14,8 +14,9 @@ in alpha. For more information about Takahē, see
    :maxdepth: 2
    :caption: Contents:
 
-   installation
    features
+   installation
+   upgrading
```
humrochagf, line 19: I put features to come first because it make more sense to showcase takahe's features before deciding on installing it, and also to have upgrading coming after installation without pushing features down in the menu.

andrewgodwin (approved):
It's a good start - I'll add in some sections for various environments (kubernetes, etc.) later on!

Issue #491, opened 2023-02-03
Unfollow don't update your follow list


humrochagf:
When you unfollow someone, you stop seeing those people's posts, but if you go to your profile and open the following list, the unfollowed profile is still there.

andrewgodwin:
Yes, as you have discovered in the PR, we were missing a filter.

PR #494, opened 2023-02-03
Fix filter to display only the ones you are actually following

humrochagf:
closes: #491

### DIFF
--- a/users/services/identity.py
+++ b/users/services/identity.py
@@ -58,7 +58,10 @@ class IdentityService:
 
     def following(self) -> models.QuerySet[Identity]:
         return (
-            Identity.objects.filter(inbound_follows__source=self.identity)
+            Identity.objects.filter(
+                inbound_follows__source=self.identity,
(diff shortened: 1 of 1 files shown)


users/services/identity.py
```diff
@@ -58,7 +58,10 @@ def admin_identities(cls) -> models.QuerySet[Identity]:
 
     def following(self) -> models.QuerySet[Identity]:
         return (
-            Identity.objects.filter(inbound_follows__source=self.identity)
+            Identity.objects.filter(
+                inbound_follows__source=self.identity,
+                inbound_follows__state=FollowStates.accepted,
```
andrewgodwin, line 63: Could you change this to be `__in=FollowStates.group_active()`? I think including pending follows makes sense.

humrochagf: Sure, just updated it :slightly_smiling_face:

andrewgodwin (approved):

andrewgodwin:
Thanks!

PR #499, opened 2023-02-05
Hatchway API Rewrite

andrewgodwin:
Removes django-ninja and replaces it with a new API framework, "hatchway".

I plan to move hatchway into its own project very soon.


vitalik:
Hi @andrewgodwin 

Could you share some thoughts on why you dropped django-ninja ?
Thank you

andrewgodwin:
@vitalik Mostly because the Mastodon API has a very weird way of handling data where it wants to be able to draw variables from either the query string or the POST form data or the JSON body, all from the same view, plus be able to implicitly convert things written as `variable[]` to arrays. I tried to make Ninja do that, but I just couldn't convince it to, as too much of the parameter parsing is done at an early stage that I couldn't merge the POST data with the query args for the things that needed it.

PR #507, opened 2023-02-14
Update _test_settings fixture to work with Django 4.1 and 4.2

manfre:
With the `STORAGES` changes landing in Django 4.2, it is no longer happy with setting the (deprecated) `STATICFILES_STORAGE = None`. This fixture change is all that is necessary to run the tests against Django 4.2a1.


andrewgodwin (approved):

andrewgodwin:
Thanks!

PR #509, opened 2023-02-14
Fix CSV export to export followers handle instead of following user handle

### LINKED ISSUE #508: Followers export contains target user
Glad to see the import/export feature! 🙂 

I've tried both exports and noticed that the exported `followers.csv` had the wrong list of user handles. Here's an example:

```csv
Account address
shuuji3@takahe.social
shuuji3@takahe.social
shuuji3@takahe.social
shuuji3@takahe.social
shuuji3@takahe.social
shuuji3@takahe.social
```

I think the `following.csv` had a correct list of users.

shuuji3:
resolves #508

Let's say I'm "User A". When "User B" follows "User A", it creates a Follow like `<Follow: #148337992523677940: userB@takahe.social → userA@takahe.social>` (source -> target)

So to create a list of my followers, I need to use `follow.source.handle` instead of `follow.target.handle` to get a follower's handle name.

### DIFF
--- a/tests/users/views/test_import_export.py
+++ b/tests/users/views/test_import_export.py
@@ -64,3 +64,23 @@ def test_export_following(
         response.content.strip()
         == b"Account address,Show boosts,Notify on new posts,Languages\r\ntest@remote.test,true,false,"
     )
+
+
+@pytest.mark.django_db
+def test_export_followers(
+    client_with_identity: Client,
+    identity: Identity,
+    identity2: Identity,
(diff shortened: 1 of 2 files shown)


andrewgodwin (approved):

andrewgodwin:
Lovely, thanks - I'll definitely take the extra test, even if I fixed the bug as soon as I saw the issue title and went "oh, I'm an idiot"!

PR #511, opened 2023-02-15
Fix boost reach

humrochagf:
Posts had the target logic, but boosts were mainly being delivered locally and to the post author.

Now they are sent to all your followers (that are following your boosts).


andrewgodwin (approved):

andrewgodwin:
Thanks!

Issue #448, opened 2023-01-19
Hashtags on our posts served from Mastodon appear as normal links


andrewgodwin:
Hashtags work fine in our web UI, on clients talking to our API, and on the Mastodon Web UI, but do not work properly on clients talking to the proper Mastodon API and retrieving our posts.

Something about our hashtags is making it through Mastodon's internal storage out into their client API, and I have no idea what.

manfre:
Most likely a result of preferring the href when stripping the post content down to plaintext.

andrewgodwin:
It's not that - it's something about how Mastodon presents hashtags to clients versus in their UI (where it totally works).

I verified that Elk also gets it wrong when seeing one of our posts on a Mastodon server, so it's presumably something we have to add onto our HTML that mastodon will then pass through and not rewrite?

andrewgodwin:
OK, I've tried adding `rel=tag` and the classes, and they don't make a difference. Someone is going to have to go into the source code of a client and figure out what's happening and why they disagree with the Mastodon web UI.

glassresistor:
this looks like it might be where elk decides if content is a hashtag
https://github.com/elk-zone/elk/blob/359cc4935b5dee59743add577a540b04a32f643c/composables/content-parse.ts#L21

andrewgodwin:
I don't think so - that just looks like a general list of classes to allow through into the HTML. I'd be expecting something to go looking in the `tags` part of the API response, maybe.

humrochagf:
Following up on the question in the MR @andrewgodwin I just ran some more tests with my instances to validate the boost case and it seems that it's all fine, we can close this issue, unless someone has found a corner case.

andrewgodwin:
This issue is specifically about them not working in certain API clients - is that where you checked?

humrochagf:
Tested on tusky and now on elk, both receiving boost and posts from takahe accounts to Mastodon (fosstodon) ones.
Looking from takahe side it will not display anything since we don't have preview card feature implemented.

![Screenshot from 2023-02-20 12-11-36](https://user-images.githubusercontent.com/1223625/220167670-c46ecba4-2af2-47d6-bde7-f9939c2a1fcc.png)

humrochagf:
Ah, and Mastodon frontend interface is also working

andrewgodwin:
Lovely! Thank you for the through testing!

PR #515, opened 2023-02-20
Fix hashtag being handled as common link in the preview card

humrochagf:
After looking into the Mastodon code I found out that it has a special treatment to identify hashtags and skip them while looking for candidates for the preview card.

To mark a link as hashtag it checks for the `rel="tag"` attribute in the links.

https://github.com/mastodon/mastodon/blob/main/app/services/fetch_link_card_service.rb#L97

Notice that once Mastodon ingests it, the `rel="tag"` gets removed from the link and the link receives a `class="status-link"` at the interface.

### DIFF
--- a/core/html.py
+++ b/core/html.py
@@ -200,9 +200,9 @@ class FediverseHtmlParser(HTMLParser):
         hashtag = hashtag.lstrip("#")
         self.hashtags.add(hashtag.lower())
         if self.uri_domain:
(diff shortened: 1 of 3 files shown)


humrochagf:
Before:
Screenshot from 2023-02-19 22-27-59

After:
Screenshot from 2023-02-19 22-27-38

andrewgodwin (approved):

andrewgodwin:
Thanks for digging into that! I wonder if this will also help fix #448.

PR #516, opened 2023-02-20
Fix nameserver substitution for IPv6 resolvers

colin:
It seems, the 0.8 release broke deployments on fly.io as the nginx restarts fails with
> nginx: [emerg] invalid port in resolver "fdaa::3" in /etc/nginx/conf.d/default.conf:77

Apparently, fly.io is using IPv6 nameservers internally and if I understand https://nginx.org/en/docs/http/ngx_http_core_module.html#resolver correctly, IPv6 addresses need to be enclosed in `[]` in the `resolver` line in the nginx config file (to differentiate them from port specifications).

### DIFF
--- a/docker/run.sh
+++ b/docker/run.sh
@@ -1,8 +1,9 @@
 #!/bin/bash
 
 # Set up cache size and nameserver subs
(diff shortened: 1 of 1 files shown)


andrewgodwin (approved):

andrewgodwin:
Thanks!

Issue #283, opened 2022-12-27
When composing a Post with images, it's too easy to skip the "Upload" step


manfre:
I keep making Posts with an image selected, but not uploaded. Need some hyperscript to notify when there is an un-uploaded attachment.

andrewgodwin:
I wonder if we should just have the browse field and description field there in the first place and just allow a single image upload? The current process is kind of janky all around.

TkTech:
I personally upload multiple images all the time, would really suck to lose it

manfre:
@TkTech I agree that multiple image support is a must have feature.

andrewgodwin:
OK, OK! Let's try to make the code a bit easier to work with, then.

whatSocks:
I know I've thought image not uploaded since I assumed they were already uploaded. 
Another nice-to-have would be a more salient warning when there's no alt text on the image, like this from hometown: 

https://github.com/mastodon/mastodon/pull/22831

andrewgodwin:
This has been solved by me just removing this whole compose image upload thing. You can now upload only 0 or 1 images, since we will push users to an app for most composing.

PR #521, opened 2023-02-22
Put no ALT badge on attached images

### LINKED ISSUE #336: Icon to mark image as having alt text or not
This isn't really a issue on desktop but visiting takahe on mobile and trying to read image alt text results in a lot of clicking of images, which opens the file (and not in a nice image viewer). If there were an icon, it would save some time and confusion. 

Some examples UIs [are discussed here](https://github.com/mastodon/mastodon/pull/22831)  

I like this one since it's unobtrusive

https://github.com/mastodon/mastodon/pull/22831#issuecomment-1367638566

shuuji3:
resolves #336

This PR adds a "No ALT" (x ALT) badge on the left bottom place of an image if it has no description.

**Example:**
Screenshot 2023-01-19 at 13 47 02

But I think there are other options too:

Screenshot 2023-01-19 at 13 46 43
Screenshot 2023-01-19 at 13 48 34
Screenshot 2023-01-19 at 13 47 30

I choose the current one because it prevents the actual image from being hidden (some people don't like that and ask some clients to add an option to hide the ALT badge) as well as it can encourage people to write a description for better accessibility.

### DIFF
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -1506,6 +1506,7 @@ form .post {
 }
 
 .post .attachments a {
+    position: relative;
     display: block;
     width: 100%;
 }
@@ -1516,6 +1517,18 @@ form .post {
     object-fit: cover;
     object-position: center;
     max-height: 400px;
+}
+
+.post .attachments .badge {
+    position: absolute;
+    bottom: 6px;
+    left: 6px;
+    padding: 0 4px;
+    border-radius: 4px;
+    font-size: 14px;
+    font-weight: bold;
+    color: var(--color-text-in-highlight);
+    background: rgb(0 0 0 / 60%);
 }
 
 .post .attachments video {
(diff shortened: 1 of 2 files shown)


tobru:
Personally, I like the last one the best (show "ALT" when there is one, showing no tag when there is no ALT). But could live with the suggested implementation as well.

shuuji3:
Thanks for your input.🙂 Yes, it also could be familiar with more people since it seems popular way to show alt badge at this time. I will leave it up to Andrew. I can update whichever pattern.

andrewgodwin:
Showing ALT with a description and not having one with no description is the current pattern in most other places I've seen, so I'd like to stick to that so we are less surprising.

I think the time to shout at users about there not being alt-text is probably in the composer or maybe when they go to boost!

shuuji3:
OK, I switched to the standard way like the last option:



Also, I remembered there was a discussion about a warning for no alt image during uploading at #283. We can improve that process for no description image.

andrewgodwin (approved):

andrewgodwin:
Thanks!

PR #525, opened 2023-02-25
Fix navigation menu in mobile

humrochagf:
Some options were unreachable and the menu was scrolling horizontally


humrochagf:
Before:
mobile-menu-before

After:
mobile-menu-after

andrewgodwin (approved):

andrewgodwin:
Thanks!

Issue #528, opened 2023-02-28
Cannot follow from GoToSocial


kvibber:
Possibly different from #223 , this doesn't seem to be namespace related.

Trying to follow a Takahe account from a GoToSocial account. Takahe doesn't register it, and GoToSocial considers the follow request pending until the user cancels it.

I have some debug info from the GoToSocial side, where it's trying to post the following JSON to https://takahe.social/inbox/ and getting a 400 "message processing error"

```
{
  "@context": "https://www.w3.org/ns/activitystreams",
  "actor": "https://gts.keysmash.xyz/users/kelsonv",
  "id": "https://gts.keysmash.xyz/users/kelsonv/follow/019WV3Y41NQFXNYN155YB3BKDA",
  "object": "https://takahe.social/@ktv@takahe.social/",
  "to": "https://takahe.social/@ktv@takahe.social/",
  "type": "Follow"
}
```

kvibber:
Probably the same issue when trying to mention a Takahe account in a GoToSocial post. I'm not sure which of these two JSON objects is sent to the Takahe inbox, but the inbox is returning a 400 again.

```
[...]

PR #529, opened 2023-03-01
Accept hs2019 in signatures (Fixes part of federation with GoToSocial)

kvibber:
Reference: https://datatracker.ietf.org/doc/html/draft-cavage-http-signatures-12 It is recommended to list "hs2019" instead of the actual signature algorith in order to obscure attack vectors.

Some fediverse servers, including the library used by GoToSocial, do this, and rejecting those signatures blocked incoming requests to the inbox.

The validation being used here can still process the signature even with the algorithm field labeled this way, so allowing it fixes this part of the incompatibility described in #528 , namely that a message from a GoToSocial user mentioning a Takahe user doesn't federate to the Takahe instance.

### DIFF
--- a/core/signatures.py
+++ b/core/signatures.py
@@ -162,7 +162,10 @@ class HttpSignature:
         # Reject unknown algorithms
         # hs2019 is used by some libraries to obfuscate the real algorithm per the spec
         # https://datatracker.ietf.org/doc/html/draft-cavage-http-signatures-12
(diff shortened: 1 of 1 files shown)


andrewgodwin (approved):

andrewgodwin:
Code change looks good - you'll need to run our linter (or manually copy its formatting from the failed test) to get past the test suite though!

andrewgodwin:
Thanks!

PR #530, opened 2023-03-06
Add admin notes field to domains

### LINKED ISSUE #259: Admin notes for domains (e.g. why it's blocked)
I would love to get a small text field, I think varchar(254) should be enough, to enter a small reason, when one blocks an instance or a user from the instance.

chdorner:
Fixes #259

Adds a notes field to the various admin views for domains, including the federation edit field.

This is my first PR, I wasn't sure what the guideline on automated tests is. I couldn't find any existing tests for these admin forms.


andrewgodwin (approved):
There are in fact basically no tests for the admin forms yet! Thus, I don't mind bringing in a small change like this without tests.

(I will get them written eventually - sooner, now the admin is mostly stable, I think)

PR #531, opened 2023-03-08
Allow to set default reply visibility

### LINKED ISSUE #256: Set default visibility for replies
The current settings of `unlisted` are really helpful. Sometimes. But I had some discussions with others, who think, that it wouldn't be possible for others to join a longer discussion, when they don't see the answers, because the original post will never be bumped up in a timeline.
I saw on the Fedilab client, that they have a special setting for this:

So I thought, we could get such a setting too? It could be identity based, like the default visibility.

chdorner:
Fixes #256 

Wasn't sure if it's worth filtering the post visibility changes on the settings page to only allow to switch between public and unlisted. I guess the only one that doesn't make much sense is local only when trying to reply to a remote status.

### DIFF
--- a/activities/views/compose.py
+++ b/activities/views/compose.py
@@ -123,7 +123,9 @@ class Compose(FormView):
             if self.reply_to:
                 initial["reply_to"] = self.reply_to.pk
                 if self.reply_to.visibility == Post.Visibilities.public:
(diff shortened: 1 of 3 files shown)


andrewgodwin (approved):

andrewgodwin:
I think it's fine to leave local only in there - I could see a use case for it!

PR #532, opened 2023-03-09
Initial support for IceCubes

dcwatson:
IceCubes (an open source iOS client) is using the v1/instance API and expecting a "polls" configuration. It also `POST`s to `v1/apps` using query parameters:

> 127.0.0.1 - - [09/Mar/2023:15:39:59 +0000] "POST /api/v1/apps?client_name=IceCubesApp&redirect_uris=icecubesapp://&scopes=read%20write%20follow%20push&website=https://github.com/Dimillian/IceCubesApp HTTP/1.1" 500 3309 "-" "IceCubesApp/1735 CFNetwork/1404.0.5 Darwin/22.3.0"

This is likely a bug on their end? But in the interest of wider client support, it was easy to accept both. Not sure your feelings on how strict the interpretation of the API should be. I've tested this on my instance (https://quiet.social) - there are other issues, but this is enough to get the basics working.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,6 +6,9 @@
 .pre-commit-config.yaml
 .venv
 /fly.*
+/static-collected
+/takahe/local_settings.py
+__pycache__/
 media
 notes.md
 venv
--- a/api/views/apps.py
+++ b/api/views/apps.py
@@ -1,28 +1,27 @@
 import secrets
 
-from hatchway import Schema, api_view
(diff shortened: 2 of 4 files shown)


andrewgodwin (approved):
Yes, Mastodon's insistence on accepting literally any kind of value for submission (query, multipart, form-data, or JSON body) is one of the reasons I had to write my own API framework - it's totally valid!

PR #534, opened 2023-03-10
Fix home timeline pagination 🤞

dcwatson:
My theory is that one or both of these is happening:

1. Filtering on `subject_post_id` OR `subject_post_interaction` may cause extra timeline events because `type` is not taken into account. These could be duplicates, or be outside the requested range, etc.
2. Ordering by `TimelineEvent.id` is not a perfect proxy for post or interaction IDs

I think using the appropriate ID based on event type is the way to go, and it seems to be working well on my (tiny) instance. Not sure if it's worth calculating/storing this `event_id` when creating timeline events.

### DIFF
--- a/api/pagination.py
+++ b/api/pagination.py
@@ -4,10 +4,11 @@ from collections.abc import Callable
 from typing import Any, Generic, Protocol, TypeVar
 
 from django.db import models
+from django.db.models.expressions import Case, F, When
 from django.http import HttpRequest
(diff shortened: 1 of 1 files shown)


andrewgodwin (approved):

andrewgodwin:
This may or may not perform badly - I am going to merge it in and then run it on takahe.social to see.

PR #539, opened 2023-03-12
Follow hashtags

chdorner:
I had some time this weekend and went ahead and implemented what I wrote down here before getting feedback on Discord. If this is completely wrong then at least it was a good learning opportunity to get to know the codebase :)

There's still an open question on how to implement the `/api/v1/followed_tags` endpoint given that it needs proper IDs for pagination, but the `Hashtag` model just uses the slug as the primary key.


users/models/hashtag_follow.py
```diff
@@ -0,0 +1,52 @@
...
+        return self.get_queryset().by_hashtags(hashtags)
+
+
+class HashtagFollow(models.Model):
+    identity = models.ForeignKey(
+        "users.Identity",
+        on_delete=models.CASCADE,
+        related_name="hashtag_follows",
+    )
+    hashtag = models.ForeignKey(
```
andrewgodwin, line 26: This is the target of that query in aget_targets, so I think it should have `db_index=True`

api/views/tags.py
```diff
@@ -14,5 +17,53 @@ def followed_tags(
     min_id: str | None = None,
     limit: int = 100,
 ) -> list[schemas.Tag]:
-    # We don't implement this yet
-    return []
+    queryset = Hashtag.objects.followed_by(request.identity)
+    paginator = MastodonPaginator()
+    # TODO: this fails due to the Hashtag model not having an `id` field
```
andrewgodwin, line 22: Maybe try paginating over the HashtagFollow objects instead?

users/models/hashtag_follow.py
chdorner: Added 👍

api/views/tags.py
chdorner: Yeah, that's a good idea. I added a new `FollowedTag` schema which adds the ID from the `HashtagFollow` model, which works well enough 👍

chdorner:
I left the WIP commit in there for easier re-review but can squash before merge

andrewgodwin (approved):

andrewgodwin:
I use "squash and merge" on this side anyway, so no need to do it. Thanks for this - a lovely bit of work!

Issue #541, opened 2023-03-14
Takahē posts viewed in clients open hashtag as URL


chdorner:
This is related to #448 but not exactly the same.

When directly logged in to Takahē with certain clients and I click on a hashtag of one of my own posts, those clients open the browser with the hashtag link.

So far I can see this happening with:
- Ivory
- Mona
- Semaphore / Pinafore
- Elk

Additional clients may be affected too.

IceCubes is the only client I saw it working correctly

andrewgodwin:
That was in fact what happened with #448 - the hashtags were parsed as links by clients.

chdorner:
Right, I understood #448 to be specifically for clients talking to a proper Mastodon API retrieving a Takahē status.

I'm seeing the same bug happening with these clients but when they talk directly to the Takahē API

andrewgodwin:
It's all about how the links are formatted - the previous fix was to have `rel="tag"` in there. If that's gone it's definitely a regression.

chdorner:
It does have `rel="tag"`

this is the JSON according to `/debug/json`:

```js
{
    // ...
    "content": "<p><a href=\"https://social.chdorner.com/tags/lightroom/\" rel=\"tag\">#Lightroom</a>, go home you&#x27;re drunk.</p>",
    "tag": [
        {
            "href": "https://social.chdorner.com/tags/lightroom/",
            "name": "#lightroom",
            "type": "Hashtag"
        }
    ],
    // ...
}
```

the only difference I can see from an example originating from Mastodon is the `class="mention hashtag"`, in i.e.: `<p><a href="https://vmst.io/tags/mastoadmin" class="mention hashtag" rel="tag">#<span>mastoadmin</span></a></p>`
from [this](https://vmst.io/@vmstan/110027699424938735) status.

chdorner:
Found the Elk parser: https://github.com/elk-zone/elk/blob/dc397e133c9009ea8f6a014cf453cc602d3c96eb/composables/content-render.ts#L83-L103

Almost seems like it's ignoring the `tags` field returned by the Mastodon API entirely.
Seems that it goes through the content field and checks all anchor tags with a class `"mention"`, then match against `/^https?:\/\/([^/]+)\/tags\/([^/]+)$/` and use the latest match group as the tag name.

Unclear how other clients implement this.

chdorner:
[...]

PR #542, opened 2023-03-15
Add `mention` and `hashtag` classes to hashtag anchors

chdorner:
Related to #541 

* This fixes hashtag detection in Ivory and Mona.
* Elk doesn't work yet because the regex fails to match due to the trailing slash of the URL
* Semaphore doesn't use a regex, but checks the suffix of the URL, so fails as well due to the trailing slash


andrewgodwin (approved):

andrewgodwin:
Yeah, I don't think we're going to be able to fix the other clients without gently encouraging them to not require a trailing slash.

PR #543, opened 2023-03-16
Always return voters_count

humrochagf:
Even though mastodon says in doc that `voters_count` can be null, it returns the `voters_count` even when it's not a multiple choice case, and some clients like the main mastodon and megalodon (that uses the same codebase) relies on it to display the totals.


andrewgodwin (approved):

andrewgodwin:
Ah yes, never rely on the Mastodon docs :D

Issue #498, opened 2023-02-05
Developer API keys / Authorised applications


chris-y:
Mastodon has a UI for adding applications, setting permissions, and providing the client key/secret/access token. I find this incredibly useful for personal apps/bots that don't need the oauth flow. Please can this be added to Takahe along with the ability to see and revoke permissions for current authorised apps.

andrewgodwin:
Yup, the whole applications UI does need writing, and I agree an easy way to make personal keys should be in there.

andrewgodwin:
This has been merged in as part of the big refactor - you can see what apps are authd, revoke their access, and make a personal access token and view it.

PR #552, opened 2023-03-31
Add Apps Settings

CSDUMMI:
This addresses #498. 

I implemented two new views:
- `/settings/apps/add` to create a new app
- `/settings/apps` is a list of all currently registered apps, their client id, secret and current access token.

It currently lacks any link to either of these pages from the settings sidebar.

This is my first contribution to this repository and my first work on a Django project.  I'll probably have missed some conventions/rules/best practices and I understand if you  reject this PR and do this better yourself.  But I'm willing to improve it if you can tell me how.

### DIFF
--- a/templates/settings/add_apps.html
+++ b/templates/settings/add_apps.html
@@ -1,0 +1,18 @@
+{% extends "settings/base.html" %}
+
+{% block subtitle %}Add an App{% endblock %}
+
+{% block content %}
+    <form action="/settings/apps/add" method="POST">
+        {% csrf_token %}
+
+        {% include "forms/_field.html" with field=form.name %}
+        {% include "forms/_field.html" with field=form.redirect_uri %}
+        {% include "forms/_field.html" with field=form.website %}
+        {% include "forms/_field.html" with field=form.scopes %}
+
+        <div class="buttons">
+            <button>Add Application</button>
+        </div>
+    </form>
+{% endblock %}
--- a/templates/settings/apps.html
+++ b/templates/settings/apps.html
@@ -1,0 +1,33 @@
+{% extends "settings/base.html" %}
+
+{% block subtitle %}Apps{% endblock %}
+
+{% block content %}
+    <h2>Apps</h2>
+    <div>
+        {% for token in object_list %}
+            <p>
+                <h3>{{token.application.name}}</h3>
+                <table class="items">
+                    <tr>
+                        <td>Client ID</td>
+                        <td>{{token.application.client_id}}</td>
+                    </tr>
+                    <tr>
+                        <td>Client Secret</td>
+                        <td>{{token.application.client_secret}}</td>
+                    </tr>
+                    <tr>
+                        <td>Access Token</td>
+                        <td>{{token.token}}</td>
+                    </tr>
+                    <tr>
+                        <td>Scopes</td>
+                        <td>{{token.scopes}}</td>
+                    </tr>
+                </table>
+            </p>
+        {% endfor %}
+    </div>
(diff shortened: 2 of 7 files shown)


templates/settings/add_apps.html
```diff
@@ -0,0 +1,13 @@
+{% extends "settings/base.html" %}
+
+{% block subtitle %}Add an App{% endblock %}
+
+{% block content %}
+    <form action="/settings/apps/add" method="POST" enctype="multipart/form-data">
+        {% csrf_token %}
+        {{ form.as_p }}
```
andrewgodwin, line 8: There's an existing form pattern style with an include per field you can use in the codebase - if you look at the other forms you'll be able to copy it.

templates/settings/add_apps.html
```diff
@@ -0,0 +1,13 @@
+{% extends "settings/base.html" %}
+
+{% block subtitle %}Add an App{% endblock %}
+
+{% block content %}
+    <form action="/settings/apps/add" method="POST" enctype="multipart/form-data">
```
andrewgodwin, line 6: This doesn't need to be multipart, there's no file uploads.

templates/settings/apps.html
```diff
@@ -0,0 +1,33 @@
...
+
+{% block subtitle %}Apps{% endblock %}
+
+{% block content %}
+    <h2>Apps</h2>
+    <div>
+        {% for token in object_list %}
+            <p>
+                <h3>{{token.application.name}}</h3>
+                <table>
```
andrewgodwin, line 11: There's existing setting page table styling, please use that! It'll allow for addition of "revoke token" actions and the like later.

users/views/settings/add_apps_page.py
```diff
@@ -0,0 +1,51 @@
...
+from django.utils.decorators import method_decorator
+from django.views.generic import FormView
+
+from api.models.application import Application
+from api.models.token import Token
+from users.decorators import identity_required
+
+
+class AppsForm(forms.Form):
+    name = forms.CharField()
```
andrewgodwin, line 14: Each of these fields should have `help_text` set.

users/views/settings/__init__.py
```diff
@@ -2,6 +2,8 @@
 from django.views.generic import RedirectView
 
 from users.decorators import identity_required
+from users.views.settings.add_apps_page import AddAppsPage
+from users.views.settings.apps_page import AppsPage
```
andrewgodwin, line 6: These should both be in one module, called `apps`.

andrewgodwin (commented):
The overall idea makes sense - the conflating of app tokens vs. make an app is probably fine. There's a lot of places where you've diverged from the existing codebase style, though - I've left you some comments where I saw it, and the linter in CI is also going to pick up on a lot of other things.

templates/settings/apps.html
CSDUMMI: Do you mean the "items" CSS class used in import_export.html?

CSDUMMI: > There's existing setting page table styling, please use that! It'll allow for addition of "revoke token" actions and the like later.

Should this "revoke token" button be added as a part of this PR?

andrewgodwin:
I was refactoring the UI anyway, so I manually wrote this in as part of that new branch - you can see that commit here: https://github.com/jointakahe/takahe/commit/32e74a620d7790282ef18013ce1147d929a5b592

It's the same rough idea as this, but with different naming (based on "token") for internal methods, doesn't leak app secrets, hides the access token behind a click-to-view, and allows revoking of access tokens.

PR #555, opened 2023-04-07
Django 4.2 with psycopg 3

manfre:
* Currently running on https://manfre.net/
* No Django version related code changes needed. Manly new deprecations that can be updated in future PRs.


pauloxnet:
Have you tried to run django-upgrade hook with pre-commit after setting Django 4.2 as target version?

manfre:
> Have you tried to run django-upgrade hook with pre-commit after setting Django 4.2 as target version?

Not yet. It likely will not work well if it attempts to auto-upgrade the the `index_together` deprecation. We're doing some not straightforward things for the Stator base model to ensure certain indexes exist.

PR #556, opened 2023-04-10
Support editing media description when editing statuses

chdorner:
The `media_attributes` key has not made it into the docs yet, but the functionality has been merged in mastodon/mastodon#20878.

At least Ivory is using this key, and it's only sent when there are changes, so we preserve alt texts when the key is not in the request.
Mona on the other hand seems to not use this, but calls `POST /api/v2/media` before editing the post.


andrewgodwin:
I didn't realise clients were already using it! Thanks.

PR #558, opened 2023-04-17
Sentry/S3 storage/Google storage should be optional

GnunuX:
I don't want to use S3/Google storage nor Sentry. I don't want install those dependencies.

I this case Takahe cannot start.


andrewgodwin:
We only support installing via the Docker image where these are already bundled, so unfortunately I'm not going to take this one.

PR #561, opened 2023-04-25
Pinned posts

chdorner:
A bit of a bigger one:

- Web UI:
  - Shows pinned posts on top of identity's profile
  - Adds pin/unpin actions to the identity's own post actions
- API:
  - Returns pinned posts in `/api/v1/accounts/<id>/statuses` when `pinned=1` is set
  - Implements `/api/v1/statuses/<id>/pin` and `/api/v1/statuses/<id>/unpin` endpoints
- Federation:
  - Makes pinned posts available under `/@<handle>/collections/featured/` and adds a link to it in the AP actor view
  - Fans out local pin and unpin actions as Add/Remove AP activities to all followers
  - Handles incoming Add/Remove AP activities from remote identities
  - Syncs all pinned posts when fetching/updating an actor

Not super happy how the pinned posts are synced when fetching an actor, but I didn't immediately find a better place other than tacking it at the end of `Identity.fetch_actor`. Also, this is the first time I did some changes to code run by stator, so I might have gotten some stuff wrong in there too!

Pinning / unpinning my own posts:
Screenshot 2023-04-25 at 15 49 12
Screenshot 2023-04-25 at 15 49 30

Pinned posts on another profile:
Screenshot 2023-04-25 at 15 45 43


chdorner:
@andrewgodwin I believe this should be working as intented now, but without the `post.features` thing yet. This could be a future improvement though, making the `/api/v1/accounts/<id>/statuses?pinned=true` query slightly more performant by not having to join on the interaction table.

Also, a note on when this is rolled out (or when you're testing it manually), a given identity needs to be refreshed first since it's `featured_collection_uri` needs to be set to correctly process Add/Remove activities.

andrewgodwin (approved):

andrewgodwin:
Thanks so much for this!

PR #562, opened 2023-05-02
Allow Link header in CORS

chdorner:
This is one of the two issues with getting pagination to work in Elk.

The other one being that masto.js (which Elk uses) is not parsing the `Link` header correctly given that it expects the `next` link to come first. [A fix](https://github.com/neet/masto.js/pull/899) has been merged, but not yet released or the dependency upgraded in Elk. We could swap those two links around in the meantime if we want pagination to work already in Elk, given the upcoming UI changes in Takahe 0.9.0.


andrewgodwin:
Yeah, I'd rather swap them over and have it work now, given I want to push the big refactor onto `main` this week.

PR #563, opened 2023-05-02
Fix post attachment author check on editing

chdorner:
Found this when adding alt text while composing a post in Elk, it sends a request to `PUT /api/v1/media/<id>` when the post attachment doesn't have a post yet.


andrewgodwin:
Should have guessed editing for non-attached attachments was a thing!

PR #565, opened 2023-05-03
UI Refactor

andrewgodwin:
Removes all timeline UI from the web server, and moves from identity-in-session to identity-in-URL for all remaining pages.


users/middleware.py
```diff
@@ -1,39 +1,17 @@
...
-                    request.identity = Identity.objects.get(id=identity_id)
-                    User.objects.filter(pk=request.user.pk).update(
-                        last_seen=timezone.now()
-                    )
-                except Identity.DoesNotExist:
-                    request.identity = None
-
+        request.domain = None
+        if "host" in request.headers:
+            request.domain = Domain.get_domain(request.headers["host"])
```
karthikb351, line 15: My django is a bit rusty, but shouldn't we be caching this in the session object and only reading from the db on a miss? (Or am I misunderstanding how this works)

andrewgodwin: Yeah, it should probably be cached _somewhere_, but it's not a massive hit to do it once per page load.

PR #568, opened 2023-05-04
fix: show domain setup to admins

karthikb351:
Prompts admin users to setup domains on the identity creation

### DIFF
--- a/users/views/identity.py
+++ b/users/views/identity.py
@@ -370,3 +370,10 @@ class CreateIdentity(FormView):
         )
         self.request.session["identity_id"] = identity.id
         return redirect(identity.urls.view)
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["user"] = self.request.user
+        if len(context["form"].fields["domain"].choices) == 0:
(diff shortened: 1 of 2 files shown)


andrewgodwin:
This looks reasonable - there's a few wording changes I'd want to do but I could easy get those in after a merge.

andrewgodwin (approved):

users/views/identity.py
```diff
@@ -370,3 +370,10 @@ def form_valid(self, form):
         )
         self.request.session["identity_id"] = identity.id
         return redirect(identity.urls.view)
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["user"] = self.request.user
+        if len(context["form"].fields["domain"].choices) == 0:
```
karthikb351, line 377: This feels a bit ugly to read from the form field choices list. I'm not familiar enough with django formviews to know where to fetch domains so it's passed to both the form class init as well as have it in context.

andrewgodwin: Yeah, it's not ideal, but not terrible - I'll merge it and then see if I can replace it with a more direct query.

PR #570, opened 2023-05-06
Migrate to new staticfiles storage setting

karthikb351:
Migrates to the new settings key for staticfiles storage.

`STATICFILES_STORAGE` was deprecreated in django 4.2, and the docs suggest moving to the `STORAGES` setting.

Docs: 
* https://django.readthedocs.io/en/stable/ref/settings.html#staticfiles-storage
* https://django.readthedocs.io/en/stable/ref/settings.html#storages


andrewgodwin:
Thanks!

PR #571, opened 2023-05-07
Add link to other Mastodon apps

shuuji3:
Hi, how about adding the link to https://joinmastodon.org/apps to help those who may like other applications to find apps? 

Screenshot 2023-05-06 at 20 05 32

I'm personally using these three applications on each platform and I think these are good choices. 🙂


andrewgodwin:
I really don't want to link to that page as it pushes the Mastodon official app as the top and main option, and that's maybe the worst thing to push people towards in our case!

shuuji3:
OK, understand. Maybe there might be a better directory website.

PR #572, opened 2023-05-07
Prevent Cloudflare from treating handles as emails

ewpratten:
A display issue occurs when a Takahē instance is hosted behind a Cloudflare proxy **and** Cloudflare's [Email Address Obfuscation](https://developers.cloudflare.com/support/more-dashboard-apps/cloudflare-scrape-shield/what-is-email-address-obfuscation/) feature is enabled.

Due to the format of handles, Cloudflare treats them as email addresses, and will cause them to display as `@[email protected]`. See below for an example:

Screenshot from 2023-05-07 12-33-31

Luckily, Cloudflare provides [a way to force-disable false positives](https://developers.cloudflare.com/support/more-dashboard-apps/cloudflare-scrape-shield/what-is-email-address-obfuscation/#prevent-cloudflare-from-obfuscating-email).

This PR wraps every instance of `@{{identity.handle}}` I could find with the "magic comments" to force-disable email protection, and stop Cloudflare from hiding handles.

### DIFF
--- a/templates/activities/home.html
+++ b/templates/activities/home.html
@@ -12,7 +12,9 @@
                 <img src="{{ identity.local_icon_url.relative }}">
                 <span class="handle">
                     {{ identity.html_name_or_handle }}
+                    <!--email_off-->
                     <small>@{{ identity.handle }}</small>
+                    <!--/email_off-->
                 </span>
                 <button class="right secondary" _="on click go to url {{ identity.urls.view }} then halt">View</button>
                 <button class="right secondary" _="on click go to url {{ identity.urls.settings }} then halt">Settings</button>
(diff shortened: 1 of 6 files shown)


karthikb351:
@ewpratten for me this behaviour was only happening on django's hx-swap/hx-get action (like clicking on the refresh button on the homepage), not on the initial page load. Were you seeing this on the first page load as well?

ewpratten:
I see this on every page load, but only on the home timeline

andrewgodwin:
Surely there must be some way to turn this off on the Cloudflare side? I really am not a fan of changing HTML output just to satisfy one commercial vendor.

ewpratten:
There is, but that would also turn off actual email obfuscation. Plus, if I
remember correctly, only higher tier customers get to pick and choose which
subdomains have the feature enabled.

andrewgodwin:
I'm going to icebox this for a while as I think about it - as far as I'm concerned this is really a Cloudflare bug, so I'm not in a huge rush to support it.

PR #573, opened 2023-05-08
Fix default storage setting and use new STORAGES

chdorner:
I'm still not super well-versed in the Django specifics, but I believe #570 has introduced a problem with how the media storage is configured.
Apparently the use of `DEFAULT_FILE_STORAGE` and `STORAGES` is mutually exclusive and Django refused to boot. Curiously, this only happened on my production setup, not in the development server

I'm using local storage on my instance, so I didn't manually test S3/GCP configs, but looking at the diff I'm assuming that this should be alright.

### DIFF
--- a/takahe/settings.py
+++ b/takahe/settings.py
@@ -307,6 +307,9 @@ STORAGES = {
     "staticfiles": {
         "BACKEND": "django.contrib.staticfiles.storage.ManifestStaticFilesStorage"
     },
+    "default": {
+        "BACKEND": "django.core.files.storage.FileSystemStorage",
+    },
 }
 
 SESSION_ENGINE = "django.contrib.sessions.backends.signed_cookies"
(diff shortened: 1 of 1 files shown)


chdorner:
@karthikb351 no worries :) just discovered it by accident since I keep my instance as up-to-date with `main` as possible

takahe/settings.py
```diff
@@ -404,14 +404,18 @@ class Config:
     parsed = urllib.parse.urlparse(SETUP.MEDIA_BACKEND)
     query = urllib.parse.parse_qs(parsed.query)
     if parsed.scheme == "gs":
-        DEFAULT_FILE_STORAGE = "core.uploads.TakaheGoogleCloudStorage"
+        STORAGES["default"] = {
+            "BACKEND": "core.uploads.TakaheGoogleCloudStorage",
+        }
```
pauloxnet, line 409: This way you don't overwrite other 'default' keys:
```suggestion
        STORAGES["default"]["BACKEND"] = "core.uploads.TakaheGoogleCloudStorage"
```
The same is valid for below changes.

chdorner: Thanks! Applied this

pauloxnet (approved):

andrewgodwin:
My thanks for fixing this before I even had a chance to run the changes locally!

PR #575, opened 2023-05-09
Add `mention` class to user mention links

chdorner:
This should fix mention links in Elk to keep linking inside Elk, not to the instance of the mentioned user.

Elk was failing to detect a user mention in this if statement: https://github.com/elk-zone/elk/blob/fc97e8ff5b5ad077417a32440a978e49b8449cc0/composables/content-render.ts#L88

I could've sworn that this was working up until recently, but none of the code responsible on both the Takahe and the Elk side has changed in months.

### DIFF
--- a/core/html.py
+++ b/core/html.py
@@ -196,7 +196,7 @@ class FediverseHtmlParser(HTMLParser):
                 self.mention_aliases[short_hash] = handle_hash
             elif self.mention_aliases.get(short_hash) != handle_hash:
                 short_handle = handle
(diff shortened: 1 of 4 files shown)


chdorner:
Github is having issues, will push the test fixes as soon as I can

andrewgodwin:
> I could've sworn that this was working up until recently, but none of the code responsible on both the Takahe and the Elk side has changed in months.

Yeah, that's happened with a few things recently, like profile HTML now not working in Mastodon. Thanks for the fix!

PR #576, opened 2023-05-10
Fix hashtag search results

chdorner:
We mistakenly wrote to the key "hashtag" instead of "hashtags", resulting in no results in the API response. Additionally, the type of the Tag's `history` needs to be a list, not a dict.

This fixes hashtag search in Elk:

image

We still don't have the history, which Elk then renders as "0 people in the past 2 days".

### DIFF
--- a/api/schemas.py
+++ b/api/schemas.py
@@ -275,7 +275,7 @@ class Notification(Schema):
 class Tag(Schema):
     name: str
     url: str
-    history: dict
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Thanks!

(Sorry I haven't got to the pinned posts one yet - these are just way easier to review in between work commitments. I'll get to it soon!)

PR #578, opened 2023-05-14
Customize gunicorn worker count

karthikb351:
Allow the gnunicorn worker count to be customize via an environment variable.

We may also want to revisit what this default is. I am running takahe on the always-free tier on GCP and needed to run only 1 worker for web and stator to make sure I stayed under the memory limit.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,3 @@
-web: gunicorn takahe.wsgi:application --workers 8
+web: gunicorn takahe.wsgi:application --workers ${TAKAHE_WORKERS:-8}
 worker: python manage.py runstator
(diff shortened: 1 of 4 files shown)


andrewgodwin:
Well, that's why GUNICORN_CMD_ARGS is already an environment variable - so you can customise the workers!

andrewgodwin:
Yes, I'd prefer just documenting the usage of that command - people may also want to change things like the type of worker class being used, and so I think having a general catch-all command is fine here.

andrewgodwin (approved):

andrewgodwin:
Thanks!

PR #579, opened 2023-05-15
Support setting language on posts

chdorner:
I _think_ this is everything that's needed to implement Mastodon's post language feature.

* New identity config setting for the preferred language, will be used when creating local posts without a `language` key in the request.
* Support for the Mastodon API, either creating/updating a post, or wherever posts are rendered
* Support for ActivityPub federation via `contentMap`


activities/models/post.py
```diff
@@ -252,6 +253,9 @@ class Types(models.TextChoices):
     # The main (HTML) content
     content = models.TextField()
 
+    # The language of the content
+    language = models.CharField(max_length=2, null=True)
```
chdorner, line 257: The max length is not currently validated anywhere. I'm filtering out the most obvious one `"und"`, but other than that this could fail.
I wonder if it's better to let it fail if something weird comes along, or if it should just swallow it and default the language to `None` when that happens,

pauloxnet:
LGTM

You can also update the [content_vector_gin](https://github.com/jointakahe/takahe/blob/main/activities/models/post.py#L316) to use the new `language` field in the `SearchVector`'s `config` attribute, as done in the djangoproject.com search document:
https://github.com/django/djangoproject.com/blob/main/docs/search.py#L42

chdorner:
@pauloxnet interesting. just to make sure that I'm getting this right since the Mastodon languages are stored as 2-char strings (ISO-639-1, `en`, `de`, etc.), we'd have to store an extra column with the search config translating these values to the known postgres search configs (looks like there are 28, plus `simple`), and then recreate the search index pointing the config to that column instead of hardcoding the english config

pauloxnet:
It seems right.

In the Django Project code I used a dictionary to map 2 characters long iso languages into language names for PostgreSQL config.

Maybe there's a similar way to map language code into config names without an additional fiepd?

andrewgodwin:
I think the problem with changing the SearchVector language is that it's embedded into the index, is it not? We can't have 20-odd indexes on the content of posts, one for each language, and doing a search query without an index for it sounds painful.

pauloxnet:
> I think the problem with changing the SearchVector language is that it's embedded into the index, is it not?

Actually, I think you can have an index based on the language stored in a column in the same table, but I'd leave the change until after this PR is merged.

chdorner:
Did a bit more research on adding the `tsvector` config to the index for this. I can get it to work with just SQL, but having a hard time to get Django to do this for me. Any help would be appreciated!

```sql
-- adding a column of type `regconfig` to store each record's tsvector config
ALTER TABLE activities_post ADD COLUMN tsvector_config regconfig DEFAULT 'simple';

-- creating the GIN index
DROP INDEX content_vector_gin;
CREATE INDEX content_vector_gin ON activities_post USING GIN (to_tsvector(tsvector_config, content));
```

pauloxnet:
The SQL code is ok for me.
What's your issue in generating that same code with the Django ORM ?
I can guess the regconfig type of column?

activities/models/post.py
andrewgodwin: Unfortunately the IETF language code spec allows quite long language codes - for example, `gsw-u-sd-chzh` is a totally valid language tag (for Zurich German). Given that, we probably need to just give this max-length 50.

users/views/settings/posting.py
```diff
@@ -15,8 +17,24 @@ class PostingPage(SettingsPage):
...
             "help_text": "If content warnings should be expanded by default (not honoured by all clients)",
         },
+        "preferred_posting_language": {
+            "title": "Default posting language",
+            "help_text": "",
+            "choices": sorted(
+                [
+                    (lang.alpha_2, lang.name)
+                    for lang in pycountry.languages
+                    if hasattr(lang, "alpha_2")
```
andrewgodwin, line 27: This is _probably_ good enough for now - I think it's how most of the other clients work - but we might need to add support for separating the various Spanish dialects in future, for example.

andrewgodwin (commented):
On the migration - let's get this landed without it, and we'll figure that out separately. The new Django index objects _should_ support this, but if it's complex or weird enough then we'll just do a raw SQL one, since we only support PostgreSQL anyway.

activities/models/post.py
```diff
@@ -1152,7 +1174,6 @@ def to_mastodon_json(self, interactions=None, bookmarks=None, identity=None):
             if isinstance(self.type_data, QuestionData)
             else None,
             "card": None,
-            "language": None,
```
AstraLuma, line 1155: Does this need to be updated to forward the stored language?

PR #580, opened 2023-05-15
Fixes various issues with pinned posts

chdorner:
Now that this has been running on my instance for a while, I'm seeing a few issues with different servers and implementations.

1. Syncing pinned items for an identity on a blocked domain fails, probably fine to just skip syncing the featured collection when an identity from a blocked domain is outdated
2. Some Mastodon featured collections either include only URLs to posts, or a mix of embedded objects and URLs. Whenever I follow the URLs, they lead to a 404, so probably safe to ignore non-dict featured collection items.
    - Example of only URLs: `curl -H "Accept: application/activity+json" https://hachyderm.io/users/tiff/collections/featured | jq .`
    - Example of mixed types: `curl -H "Accept: application/activity+json" https://octodon.social/users/Eramdam/collections/featured | jq .`
3. Friendica featured collections contain `Create` activities, not the `Notes` by themselves. Mastodon seems to be skipping those, I've opted for importing those
    - Example: `curl -H "Accept: application/activity+json" https://art1sec.uber.space/featured/martin | jq .`

### DIFF
--- a/users/models/identity.py
+++ b/users/models/identity.py
@@ -760,15 +760,22 @@ class Identity(StatorModel):
 
         try:
             data = canonicalise(response.json(), include_security=True)
+            items: list[dict | str] = []
             if "orderedItems" in data:
-                return [
-                    item["id"]
-                    for item in reversed(data["orderedItems"])
-                    if isinstance(item, dict)
-                ]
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Thanks!

PR #581, opened 2023-05-15
Fixes various issues with pinned posts - continued

chdorner:
Found another 2 problems:

1. Sometimes the object URI of an item in a featured collections results in a 404, I've seen this either with Mastodon (presumably more of an access issue rather than actual 404?) or with some homegrown-ish blogging platform. I thought it's safe to just ignore those and allow the identity to transition from outdated to updated
    - since we use the `Post.DoesNotExist` also for other errors, this will also ignore those, i.e. SSL errors, >500 response codes, etc.
2. Friendica doesn't just have `Create` activities in featured collections, but also `Update`. I hope there won't be additional ones

I'll keep an eye on Sentry in the next few days and raise additional PRs if I see more errors coming in.

### DIFF
--- a/users/models/identity.py
+++ b/users/models/identity.py
@@ -771,7 +771,7 @@ class Identity(StatorModel):
                 if not isinstance(item, dict):
                     continue
                 post_obj: dict | None = item
-                if item["type"] == "Create":
+                if item["type"] in ["Create", "Update"]:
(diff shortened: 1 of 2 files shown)


andrewgodwin (approved):

andrewgodwin:
👍

PR #583, opened 2023-05-19
Use post id to generate summary class

karthikb351:
Use the post's id to uniquely identify the summary class.

Currently we hash the name of the summary to generate the unique css class name, which causes all summaries with the same name to expand and close together (was this behaviour intentional?)

Example: https://fedi.karthikbalakrishnan.com/@me@karthikbalakrishnan.com/posts/182274944255353776/

https://github.com/jointakahe/takahe/assets/3104454/287501d3-6875-45f8-8c2c-164e78ba76c7

### DIFF
--- a/activities/models/post.py
+++ b/activities/models/post.py
@@ -1,5 +1,4 @@
 import datetime
-import hashlib
 import json
 import mimetypes
 import ssl
--- a/activities/models/post.py
+++ b/activities/models/post.py
@@ -436,7 +436,7 @@ class Post(StatorModel):
         """
         if not self.summary:
             return ""
-        return "summary-" + hashlib.md5(self.summary.encode("utf8")).hexdigest()
+        return "summary-{self.id}"
 
     @property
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Yes, this was an entirely intentional feature - the reason being that if you are willing to expand one content warning, you may as well expand them all.

I'm curious if there are reasons _not_ to do this that aren't "it's not what most other clients do"?

karthikb351:
> I'm curious if there are reasons _not_ to do this that aren't "it's not what most other clients do"?

@andrewgodwin I think the biggest issue is that in a long thread, it ends up shifting a lot of the content up/down, such that the toggle button also moves underneath your cursor. If we only did one at a time, the button would stay in the same place.

andrewgodwin:
Yeah, that's a good point - we aren't going to sync the viewpoint. Plus, I am always a fan of the Principle of Least Surprise, which here probably includes "doing what everyone else is".

PR #585, opened 2023-05-24
Prevent dropping ellipsis URL

shuuji3:
fixes #584

This issue seems to occur because Elk uses the textContent of the Mastodon HTML as a draft text (ref. [1](https://github.com/elk-zone/elk/blob/main/components/status/StatusActionsMore.vue#L110-L110), [2](https://github.com/elk-zone/elk/blob/main/composables/content-parse.ts#L145), [3](https://github.com/elk-zone/elk/blob/main/composables/masto/statusDrafts.ts#L61), [4](https://github.com/elk-zone/elk/blob/main/composables/content-parse.ts#L76-L76)) and we are dropping the remaining URL string after 30 characters.

Here is an example `content` comparisons between Mastodon (first) and Takahē API (second):

```diff
-    "content": "<p>Takahē - return to the wild - <a href=\"https://www.youtube.com/watch?v=IG423K3pmQI\" target=\"_blank\" rel=\"nofollow noopener noreferrer\"><span class=\"invisible\">https://www.</span><span class=\"ellipsis\">youtube.com/watch?v=IG423K3pmQ</span><span class=\"invisible\">I</span></a></p>",
+    "content": "<p>Takahē - return to the wild - <a href=\"https://www.youtube.com/watch?v=IG423K3pmQI\" rel=\"nofollow\" class=\"ellipsis\" title=\"www.youtube.com/watch?v=IG423K\">www.youtube.com/watch?v=IG423K</a></p>",
```

To emulate Mastodon's link in `content`, this wraps the URL text with `<span class="ellipsis">` and `<span class="invisible">`.

I kept the `class="ellipsis"` of the `<a>` tag untouched as I don't fully understand how it affects other parts of logic yet.

Screenshot 2023-05-24 at 21 55 54

https://takahe.shuuji3.xyz/@shuuji3@takahe.shuuji3.xyz/posts/184178424987558872/

Screenshot 2023-05-24 at 21 56 59
Screenshot 2023-05-24 at 21 58 06

https://elk.zone/takahe.shuuji3.xyz/@shuuji3/184178424987558872

PR #586, opened 2023-05-26
Prevent dropping ellipsis URL

shuuji3:
Hello, this is a follow-up PR for #585.

I focused on the ellipsis part of the URL in the last PR and missed including the first protocol part like `https://` as an invisible text. Without it, Elk still failed to recover the complete URL after saving the edited post (for example, `https://takahe.social` became a plain text `takahe.social`).

Now:
Screenshot 2023-05-26 at 16 54 26


andrewgodwin (approved):

andrewgodwin:
Thanks!

Issue #285, opened 2022-12-27
Admin redirect fixes


kromonos:
When editing on specific pages, like from the Federated tab or in settings, federation, you always get redirected to the start page or, in case of the federation settings, to the first page. It would be really nice, if you get back to the page you're coming from. 
E.g. current behavior:
settings ⇾ federation ⇾ scroll until e.g. page 5 ⇾ edit a domain ⇾ federation page 1

Would be nice:
settings ⇾ federation ⇾ scroll until e.g. page 5 ⇾ edit a domain ⇾ federation page 5

PR #587, opened 2023-06-13
Bugfix admin redirect fixes

mj2p:
fixes jointakahe/takahe#285

Return to the correct page on a paginated list in the admin section after visiting a detail page it contains.

1. Set the list page index as a GET parameter in the href used to link to the detail page.
2. Capture the list page index in the `View` when creating the context
3. Set the list page index from the context as a GET parameter on the `back` link on the detail page

Works for:
- [x]  announcements
- [x]  federation
- [x]  hashtags
- [x]  identities
- [x]  invites
- [x]  reports
- [x]  users

### DIFF
--- a/templates/admin/announcement_edit.html
+++ b/templates/admin/announcement_edit.html
@@ -16,7 +16,7 @@
             {% include "forms/_field.html" with field=form.end %}
         </fieldset>
         <div class="buttons">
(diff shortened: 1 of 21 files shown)


andrewgodwin (approved):

andrewgodwin:
Thanks!

PR #589, opened 2023-06-22
Fix metadata not being updated in Mastodon instances

humrochagf:
- Adds missing fan out process when we update the account using the API (mastodon clients case)
- Use type expected by mastodon

### DIFF
--- a/users/models/identity.py
+++ b/users/models/identity.py
@@ -534,7 +534,7 @@ class Identity(StatorModel):
         if self.metadata:
             response["attachment"] = [
                 {
-                    "type": "http://schema.org#PropertyValue",
+                    "type": "PropertyValue",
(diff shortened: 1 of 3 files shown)


users/models/identity.py
```diff
@@ -534,7 +534,7 @@ def to_ap(self):
         if self.metadata:
             response["attachment"] = [
                 {
-                    "type": "http://schema.org#PropertyValue",
```
humrochagf, line 537: Took some time to understand that Mastodon sends and expects to receive `PropertyValue` instead of `http://schema.org#PropertyValue` because the original payload is hidden behind the call to `canonicalize` that expands the original value.

andrewgodwin (approved):

andrewgodwin:
Thanks!

PR #590, opened 2023-06-24
Fix canonicalize

humrochagf:
- Adds schema.org context to support gotosocial that mentions it directly (its context is huge, I only took the property value part that is the one used by attachment)
- Make canonicalize to use the json-ld context if provided instead of overriding it with a default one, this fixes most of the parsing errors that are happening during the validation process of the actors.

### DIFF
--- a/core/ld.py
+++ b/core/ld.py
@@ -592,24 +592,32 @@ def canonicalise(json_data: dict, include_security: bool = False) -> dict:
     """
     if not isinstance(json_data, dict):
         raise ValueError("Pass decoded JSON data into LDDocument")
-    context = [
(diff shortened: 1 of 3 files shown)


andrewgodwin (approved):
I'll have to test this and ensure it works fully in production, canonicalise took a lot of effort to even get to where it is now!

humrochagf:
yup, I'm running it in my instance. No issues for the moment, but its a single user personal instance. It will be nice to run on different scenarios to make sure it cover all cases 🙂

andrewgodwin:
It's been running on takahe.social now for a couple of hours so I think it's holding up nicely - still seeing things come in.

PR #591, opened 2023-06-24
Be quieter about remote hosts with invalid SSL certs

osmaa:
I was getting a lot of extremely verbose log messages and sentry errors about SSLErrors from remote hosts. Since I can't do anything about a bad certificate on someone else's server, and the SSLError anyway prevented any processing with those hosts, this patch replaces the error with a one-liner log message:

` 2023-06-24T19:32:44.009595000Z https://some.skrolli.fi/users/skrollilehti [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: certificate has expired (_ssl.c:1002)`

PR #595, opened 2023-07-03
Resubmit: Be quieter about remote hosts with invalid SSL certs

osmaa:
Resubmitting a somewhat cleaned up version #591 

I was getting a lot of extremely verbose log messages and sentry errors about SSLErrors from remote hosts. Since I can't do anything about a bad certificate on someone else's server, and the SSLError anyway prevented any processing with those hosts, this patch replaces the error with a one-liner log message:

 2023-06-24T19:32:44.009595000Z https://some.skrolli.fi/users/skrollilehti [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: certificate has expired (_ssl.c:1002)


osmaa:
well, I don't understand what the pre-commit checks are complaining about.

manfre:
This project uses pre-commit (`pip install  pre-commit`) to ensure certain code styling and validation rules are applied. `pre-commit install` will ensure the githooks are installed and run whenever you commit code. You can manually run them with`pre-commit run --all-files` and that should auto-fix the files.

andrewgodwin:
Yeah, there's an enforced code style, import sort and a few other things, but the right move is to have pre-commit run everything for you as it can fix _most_ of it automatically.

osmaa:
after being offline for a while, I figured out the pre-commit checks (thanks for the pointers!), should be good now.

andrewgodwin (approved):

Issue #592, opened 2023-06-26
IntegrityError when fetching a remote identity with pinned posts and/or with post interactions


osmaa:
My instance (which I just rebooted with a blank database due to a mishap, oops -- this may be related, since the database content doesn't match presumed federation state) where the stack traces first have a lot of DoesNotExist, but then during processing a psycopg.errors.UniqueViolation and/or django.db.utils.IntegrityError happens due to constraint violation on index "activities_post_object_uri_key".

If I'm reading this correctly, the root cause is that [actitivites.models.Post.by_ap() fetches the identity](https://github.com/jointakahe/takahe/blob/0.9.0/activities/models/post.py#L841), (including pinned posts etc?), and immediately afterwards in the same transaction a blank post with the same id is [created again](https://github.com/jointakahe/takahe/blob/0.9.0/activities/models/post.py#L853).

I have an [ugly patch for this](https://github.com/osmaa/takahe/commit/c1932b301db8d3f04548daad02b98aa335debe0f), but I'm not opening a PR with it as it's probably not the correct fix.

osmaa:
#596 and ActivityPubFormatError apparently are not enough to fully solve pinned posts causing integrity violations. My latest stack trace on this issue below. I have not yet fully parsed this, posting for reference.

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/asgiref/sync.py", line 349, in main_wrap
    raise exc_info[1]
  File "/takahe/activities/models/post.py", line 996, in by_object_uri
    return cls.objects.get(object_uri=object_uri)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/db/models/query.py", line 637, in get
    raise self.model.DoesNotExist(
activities.models.post.Post.DoesNotExist: Post matching query does not exist.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/takahe/stator/models.py", line 201, in transition_attempt
    next_state = async_to_sync(current_state.handler)(self)
[...]

PR #596, opened 2023-07-03
Fix inbox processing errors from pinned posts and non-Mastodon servers

osmaa:
If a post (interaction) comes in from AP inbox but no local author profile exists, fetch_actor will pull in both the identity AND its pinned posts, which the incoming post might have been. This would case a database integrity violation. We check for post existing again after syncing the actor.

Post processing also barfed on posts where content didn't follow Mastodon specs. For example, Kbin sets tag names in 'tag' attribute, instead of 'name' attribute.


andrewgodwin:
This is presumably the fix for #592? Code looks good, you just need to run the pre-commit hooks locally in order to get the imports sorted correctly and a few other things (see https://docs.jointakahe.org/en/latest/contributing/#coding-guidelines)

osmaa:
after being offline for a while, I figured out the pre-commit checks (thanks for the pointers!), should be good now.

andrewgodwin (approved):

andrewgodwin:
Lovely, I appreciate it!

I've also added a new way of moving inbox messages to an `errored` state if you raise an `ActivityPubFormatError` - I'm not sure that anything here warrants it, but mentioning it in case you think something could benefit from that going in later.

osmaa:
right, that change added a different way of dealing with attachments missing url. Removing mine in a separate PR.

PR #597, opened 2023-07-03
Use 400 and 401 error codes as OAuth2 documents, accept 400 as webfinger error code

osmaa:


andrewgodwin (approved):
Looks good code wise, just needs pre-commit run locally to make the linter happy!

andrewgodwin (approved):

PR #600, opened 2023-07-06
Permit SMTP to mail relay without authentication

futzle:
My Takahē installation has a working mail transport agent (exim4) right on the VM. This MTA's only purpose is to relay emails from the stator container to the world, so it does not need, or indeed accept, authentication. (Yes, it is properly firewalled and listens only on the bridge interface to the Docker container.)

The accompanying patch allows Takahē to attempt mail submission without trying to authenticate. The MTA then forwards the mail to the final recipient using whatever method it is configured to.

Tested on a Debian 11 VM.


andrewgodwin:
Looks reasonable - just needs the linting errors fixed (if you install pre-commit and then run `pre-commit run -a` it'll do the changes for you, though it looks like you could just run `black` on the file here to fix it)

andrewgodwin:
Ah, now the type checker wants you to add some types (it is almost certainly better to do that testing locally)

futzle:
Lol, spot the Perl programmer who last touched Python when it was at like 3.2.something. I'll mess around and go learn how Python does typing these days.

andrewgodwin:
If you don't want to run the pre-commit hooks directly, you just need to install and run `mypy` on the file in question and it should give you the same error - and then I suspect you just need to type that variable at its first declaration as `str | None`, since your expression can return a None value as well as a string.

futzle:
A slightly different tack: is it acceptable stylistically to change the assignment to something like:
```
if parsed.username is not None:
  EMAIL_HOST_USER = urllib.parse.unquote(parsed.username)
```
and not assign to it at all otherwise? Unset EMAIL_HOST_USER seems to be supported by the [Django email module](https://docs.djangoproject.com/en/4.2/topics/email/). Wrapping in an if-test seems to be the style of the rest of settings.py, now that I look at it.

andrewgodwin:
Yeah, that seems fine to me!

futzle:
Tested just not setting EMAIL_HOST_USER and EMAIL_HOST_PASSWORD and it can still send unauthenticated SMTP. pre-commit seems to run clean for me.

andrewgodwin (approved):

andrewgodwin:
Thanks, and I appreciate the work to iterate it and get it to pass all the linting!

futzle:
Big thanks for taking the time to get me up to speed with the toolchain and the steps I needed to take. It was a pleasant welcome!

PR #605, opened 2023-07-09
Add support to dismiss notifications

humrochagf:


activities/models/timeline_event.py
```diff
@@ -55,6 +55,7 @@ class Types(models.TextChoices):
 
     published = models.DateTimeField(default=timezone.now)
     seen = models.BooleanField(default=False)
+    dismissed = models.BooleanField(default=False)
```
andrewgodwin, line 58: We might need to consider an index on this in future, but for now I think it's selective enough it should be fine.

users/migrations/0021_remove_domain_ix_domain_state_next.py
```diff
@@ -0,0 +1,17 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("users", "0020_alter_identity_local"),
+    ]
+
+    operations = [
+        migrations.RemoveIndex(
+            model_name="domain",
+            name="ix_domain_state_next",
```
andrewgodwin, line 15: This migration should not be here - please delete the migration file (the most recent `main` commit fixes the bug that makes this appear)

humrochagf: Oups, forgot to leave it out the commit 😜 
Done

andrewgodwin (approved):

andrewgodwin:
Thanks!

Issue #613, opened 2023-07-19
Stator is inadvertently communicating with defederated domains?


futzle:
As a member of a visible minority, the first thing I did with my instance was to insert the 400 or so instances on the Oliphant Tier0 list into the users.domain model as Blocked. I consider this SOP for any Fedi instance deployment. (Side note: it'd be nice if I could import block lists on the web interface, or even to pre-emptively add blocked domains without messing with Django scripting.)

Yet I think my instance is still announcing itself to those blocked instances when the stator tries to fetch instance information from it.

```
$ sudo docker logs -t worker 2>&1 | grep genderheretics
2023-07-06T05:32:58.685628856Z Attempting transition on users.domain#genderheretics.xyz from state outdated
2023-07-07T05:33:06.614091697Z Attempting transition on users.domain#genderheretics.xyz from state updated
2023-07-07T05:33:07.195333279Z Attempting transition on users.domain#genderheretics.xyz from state outdated
2023-07-08T05:33:17.637618935Z Attempting transition on users.domain#genderheretics.xyz from state updated
2023-07-08T05:33:19.152528304Z Attempting transition on users.domain#genderheretics.xyz from state outdated
2023-07-09T05:33:27.694674737Z Attempting transition on users.domain#genderheretics.xyz from state updated
2023-07-09T05:33:28.264756791Z Attempting transition on users.domain#genderheretics.xyz from state outdated
2023-07-10T05:33:44.685896184Z Attempting transition on users.domain#genderheretics.xyz from state updated
2023-07-10T05:33:46.373056364Z Attempting transition on users.domain#genderheretics.xyz from state outdated
2023-07-11T05:34:11.901867475Z Attempting transition on users.domain#genderheretics.xyz from state updated
2023-07-11T05:34:12.429001795Z Attempting transition on users.domain#genderheretics.xyz from state outdated
```

This feels less than ideal: if the admin of that instance trawls through their webserver logs (they do), they're going to find periodic proof that my server exists.

I wonder if there's an argument for not attempting a state transition on users.Domain objects that have blocked = true, or having a dedicated `blocked` state so that our instance doesn't try to talk to that server.

What does everyone think?

andrewgodwin:
Ah yes, you're totally right, we should not be doing outgoing requests to them at all - I must have overlooked the "is the remote server alive" check when adding the blocking code.

[...]

PR #617, opened 2023-07-22
Add support to import Blocklist

humrochagf:
First iteration of blocklist support

- Adds support to simple domain list file (one domain per line)
- Adds support to Oliphant blocklist format (https://writer.oliphant.social/oliphant/the-oliphant-social-blocklist)

In this version it blocks all domains in the list since we don't have yet support to globally silencing domains.

relates to #613

### DIFF
--- a/users/views/admin/federation.py
+++ b/users/views/admin/federation.py
@@ -1,6 +1,7 @@
 import csv
 
 from django import forms
+from django.contrib import messages
 from django.core.validators import FileExtensionValidator, ValidationError
 from django.db import models
 from django.shortcuts import get_object_or_404, redirect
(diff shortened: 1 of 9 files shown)


users/views/admin/federation.py
```diff
@@ -78,3 +81,58 @@ def get_initial(self):
...
+                try:
+                    validator(domain)
+                except ValidationError:
+                    # skip adding invalid domain
+                    # to the blocklist
+                    continue
+
+                domains.append(domain)
+        except (TypeError, ValueError):
+            return redirect(".?bad_format=true")
```
andrewgodwin, line 124: You can now use the Django messages framework for this instead - `from django.contrib import messages; messages.error(request, "Your CSV was...")`. I switched some of the other error messages so far but not all.

andrewgodwin (commented):
Looks good overall, just one change to the way messages are shown if you could!

users/views/admin/federation.py
humrochagf: Done :slightly_smiling_face:

andrewgodwin (approved):

andrewgodwin:
Thanks for you work on this!

PR #618, opened 2023-07-24
add Idempotency-Key to allowed CORS header

alphatownsman:
It's used by [Mastodon](https://docs.joinmastodon.org/methods/statuses/) and Pleroma 
> Idempotency-Key
> Provide this header with any arbitrary string to prevent duplicate submissions of the same status. Consider using a hash or UUID generated client-side. Idempotency keys are stored for up to 1 hour.

Some web client will always send it regardless of server software, so I think it's reasonable to at least accept it even not implement it yet.

### DIFF
--- a/takahe/settings.py
+++ b/takahe/settings.py
@@ -9,6 +9,7 @@ import dj_database_url
 import django_cache_url
 import httpx
 import sentry_sdk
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Makes sense - thanks!

Issue #620, opened 2023-07-25
Blocked Domains trying to transition from updated to updated


humrochagf:
Hello, I noticed that now that we don't try to comunicate with blocked domains aac75dd4c3f9a5d4e17ab590c5dae421cc7c6591 and 83b57a0998d60fa3edef1d9395cda36b452f8e60 the blocked domains are trying to go from outdated to outdated and failing during this transition.

```
takahe-stator-1  | Traceback (most recent call last):
takahe-stator-1  |   File "/takahe/stator/runner.py", line 238, in clean_tasks
takahe-stator-1  |     task.result()
takahe-stator-1  |   File "/usr/local/lib/python3.11/concurrent/futures/_base.py", line 449, in result
takahe-stator-1  |     return self.__get_result()
takahe-stator-1  |            ^^^^^^^^^^^^^^^^^^^
takahe-stator-1  |   File "/usr/local/lib/python3.11/concurrent/futures/_base.py", line 401, in __get_result
takahe-stator-1  |     raise self._exception
takahe-stator-1  |   File "/usr/local/lib/python3.11/concurrent/futures/thread.py", line 58, in run
takahe-stator-1  |     result = self.fn(*self.args, **self.kwargs)
takahe-stator-1  |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
takahe-stator-1  |   File "/takahe/stator/runner.py", line 268, in task_transition
takahe-stator-1  |     result = instance.transition_attempt()
takahe-stator-1  |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
takahe-stator-1  |   File "/takahe/stator/models.py", line 216, in transition_attempt
takahe-stator-1  |     raise ValueError(
takahe-stator-1  | ValueError: Cannot transition users.domain: blob.cat: from updated to updated - not a declared transition
```

Btw, I edited the error message code locally to show the target of the failure.

I'm not yet super familiar with the underlying stator code, so I'm not sure if this is caused by a race condition or other issue but I noticed that checking the domains after the error its state keeps as updated, but the `state_next_attempt` is not incremented so it keeps failing again periodically.

andrewgodwin:
Actually, I think the right solution here is to allow that transition - I deliberately didn't want to move to outdated so that someone reading the logs might think we were communicating. Doing the transition will reset the timer to 7 days again, as well, so it shouldn't busy loop.

humrochagf:
[...]

PR #621, opened 2023-07-25
Allow updated to updated transition on Domain model

humrochagf:
This fixes the error of a domain trying to transition from updated to updated by allowing the transition on DomainStates.

Closes: #620


andrewgodwin (approved):

andrewgodwin:
Thanks!

PR #625, opened 2023-08-04
support follow request

alphatownsman:
- [x] read/write `locked` field for identity in API
 - [x] list/accept/reject follow_request for identity in API
 - [x] send accept/reject to remote server
 - [x] handle accept/reject from remote server


users/migrations/0022_follow_request.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("users", "0021_identity_aliases"),
+    ]
+
+    operations = [
+        migrations.RunSQL("DELETE FROM users_follow WHERE state <> 'accepted';"),
```
andrewgodwin, line 13: This isn't safe - you'll nuke all in-progress follow handshakes that are going on. What are you trying to achieve here?

andrewgodwin: I want to ensure that any follows that are in-flight at the time of the migration (i.e. in the `local_requested` or `remote_requested` state) will still follow through and finish.

andrewgodwin (approved):

andrewgodwin:
Great, thanks for those updates - looks good!

PR #628, opened 2023-08-06
nodeinfo reveals metadata based on domain requested

alphatownsman:
nodeName is not supported by Mastodon but other major implementations like Pleroma, Misskey and their forks all have it.

### DIFF
--- a/users/views/activitypub.py
+++ b/users/views/activitypub.py
@@ -74,8 +74,12 @@ class NodeInfo2(View):
         # Fetch some user stats
         if request.domain:
             domain_config = Config.load_domain(request.domain)
-            local_identities = Identity.objects.filter(domain=request.domain).count()
-            local_posts = Post.objects.filter(author__domain=request.domain).count()
+            local_identities = Identity.objects.filter(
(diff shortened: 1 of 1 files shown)


andrewgodwin:
I might suggest adding local into the filter query too, so the db can use that to grab the initial rows and then the scan for domain values is a lot cheaper?

TkTech:
> Thanks. very good point. I somehow had a wrong assumption that foreign key implies index but i just checked psql doc and I'm clearly wrong!

No, you're correct, but that's a behaviour of Django's `ForeignKey` and not something postgres is doing.

> A database index is automatically created on the ForeignKey. You can disable this by setting [db_index](https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.Field.db_index) to False. You may want to avoid the overhead of an index if you are creating a foreign key for consistency rather than joins, or if you will be creating an alternative index like a partial or multiple column index.

andrewgodwin (approved):
Well it makes me feel better which means it achieves something :)

Thanks for this!

PR #629, opened 2023-08-07
fix like/boost remote post

alphatownsman:
looks like there's a typo prevent like/boost remote post from working properly.


andrewgodwin (approved):

andrewgodwin:
Thanks so much for figuring this out!

PR #631, opened 2023-08-16
Add stub API endpoint for user suggestion (api/v2/suggestions)

shuuji3:
I'm testing Takahē with Elk now and fount "For you" tab always returns an error.

For you | Explore | Elk - https://elk.zone/takahe.social/explore/users
Screenshot 2023-08-16 at 13 28 16

I'm not sure if Takahē will have this functionality or not yet, but I think returning an empty response would be nice to have to avoid errors for now same as the existing trends endpoint.

### DIFF
--- a/api/urls.py
+++ b/api/urls.py
@@ -18,6 +18,7 @@ from api.views import (
     push,
     search,
     statuses,
+    suggestions,
     tags,
     timelines,
     trends,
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Yup, adding empty responses is defintely better than erroring - thanks!

PR #632, opened 2023-08-17
Fix Accept object id for follow activity for Misskey and Firefish

### LINKED ISSUE #323: Misskey accounts unable to follow
I have 2 Misskey instances and was able to locate and request to follow both of my accounts, but neither was able to successfully follow back. In Takahē I can see that I am following them and they are following me, but in Misskey it shows that following the Takahē account is still pending - it appears communication isn't getting back to Misskey to tell it that the follow successfully went through.

shuuji3:
fix #323

Yes, adjusting the `id` of `Accept` object was enough to fix this issue.

Do you know the meaning of the hash part? I couldn't find an explanation in ActivityPub spec so I just tried mimicking Mastodon's `Accept` object and it seems to work (I haven't tested yet but maybe just `#accept` could be ok).

Example object is now like this:

```diff
 {
   'type': 'Accept'
-  'id': 'https://misskey.io/follows/9iiw862tea#accept',
+  'id': 'https://takahe.shuuji3.xyz/@shuuji3@takahe.shuuji3.xyz/#accept/follows/',
   'actor': 'https://takahe.shuuji3.xyz/@shuuji3@takahe.shuuji3.xyz/',
   'object': {
     'type': 'Follow',
     'id': 'https://misskey.io/follows/9iiw862tea',
     'actor': 'https://misskey.io/users/97swdg10sa',
     'object': 'https://takahe.shuuji3.xyz/@shuuji3@takahe.shuuji3.xyz/'
   }
 }
```

I confirmed that Misskey now can follow the Takahē account:
Screenshot 2023-08-18 at 2 32 34

Also, Firefish too:
Screenshot 2023-08-18 at 1 14 29

They can see replies and new posts from Takahē after the successful following.

### DIFF
--- a/users/models/follow.py
+++ b/users/models/follow.py
@@ -230,7 +230,7 @@ class Follow(StatorModel):
         """
         return {
             "type": "Accept",
-            "id": self.uri + "#accept",
+            "id": self.target.actor_uri + "#accept/follows/",
             "actor": self.target.actor_uri,
             "object": self.to_ap(),
         }
--- a/users/models/follow.py
+++ b/users/models/follow.py
@@ -230,7 +230,7 @@ class Follow(StatorModel):
         """
         return {
             "type": "Accept",
-            "id": self.target.actor_uri + "#accept/follows/",
+            "id": f"{self.target.actor_uri}#accept/{self.id}",
             "actor": self.target.actor_uri,
             "object": self.to_ap(),
         }


andrewgodwin:
The hash part is just a way of adding ID information - previously, we were using it to add "accept" to the follow.

In your patch, though, the ID is no longer unique per follow, which it needs to be - you should shove the ID in there. Maybe add `#accept/{self.id}`?

shuuji3:
Ah, I see, and the uniqueness relied on the follow request ID previously. Then `#accept/{self.id}` should give us uniqueness.

I tested the following again and saw no issue on both servers.

andrewgodwin (approved):

andrewgodwin:
Yup, I reserved 4 bits of the snowflake IDs for type information so we can mix them in responses/pagination if needed!

Thanks for this - I do wonder if we're doing this elsewhere and ruining other compatibility with Misskey, etc.

shuuji3:
If there's a similar problem with Misskey, I'll check this issue first now.😄 

After quick tests, all basic functions such as following, being followed, posting to the timeline, viewing posts on the account page, and sending posts with image/emoji, look to work as expected.

andrewgodwin:
Great, thanks. Strange that this was the one thing that they check!

PR #633, opened 2023-08-20
signatures need to use UTF-8 in order to represent all URLs

osmaa:
Noticed this as an error pulling some signed requests as URLs with international character sets couldn't be encoded. Having reviewed Mastodon implementation, I believe it is correct to encode URLs using UTF-8. This changes nothing for majority of content since ASCII charset encodes the same way.


andrewgodwin (approved):
Thanks!

PR #634, opened 2023-08-20
fetch_account/sync_pins/post worker parallelism fix

osmaa:
After a failure in my previous attempt to eliminate IntegrityErrors following a post.DoesNotExist, a different approach. I've been running this code for a few days now and errors are gone.

The problem: an incoming post or post_interaction (boosts in particular) may call fetch_actor, which calls sync_pins, which loads posts. The whole process may take a while due to multiple remote requests, and parallel workers may be running into trying to create the same content, and one or more of the workers will then try to insert already committed post another time into the database.

Here, we 1) decouple identity commits from post commits, in order to reduce the size of the database transaction into atomic content 2) revert earlier attempt to trying to rely on the parallel commit, since transaction isolation prevented that in the database, anyway.

### DIFF
--- a/activities/models/post.py
+++ b/activities/models/post.py
@@ -834,25 +834,20 @@ class Post(StatorModel):
                 # If the author is not fetched yet, try again later
                 if author.domain is None:
                     if fetch_author:
-                        author.fetch_actor()
-                        # perhaps the entire "try again" logic below
-                        # could be replaced with TryAgainLater for
-                        # _all_ fetches, to let it handle pinned posts?
-                        if author.domain is None:
+                        if not author.fetch_actor() or author.domain is None:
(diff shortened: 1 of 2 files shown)


andrewgodwin (approved):

andrewgodwin:
Thanks for your work on this - I think that long term, I should probably move every step of this to its own stator loop (I probably want to let it be possible to make a post stub with its url and let the fetch happen async), but that'll be a while!

osmaa:
> move every step of this to its own stator loop

Would be great to be able to push an async job to the work queue from another job, I think. But yeah, this seems to run error-free now. The trace shows database SAVEPOINTs being set and cleared during the process as expected. Thank you for your patience as I was looking for a fix.

PR #636, opened 2023-08-21
do not print "Scheduling 0 handled" 

alphatownsman:
unless settings.DEBUG is on, so that it's less noisy for less busy nodes.

I personally prefer not to print it at all regardless of DEBUG, but assume it was there for a reason?


andrewgodwin (approved):

andrewgodwin:
It was mostly there so you could tell if Stator was alive when there were more async issues - I agree it makes sense to flag it behind DEBUG!

PR #639, opened 2023-08-25
ignore unknown tag type in incoming post, rather than raise exception

alphatownsman:
Various ActivityPub implementations and proposals introduced tag types, e.g. `Edition` from Bookwyrm and `Link` in [fep-e232 Object Links](https://codeberg.org/fediverse/fep/src/branch/main/fep/e232/fep-e232.md), and assume this list will continue to grow.

Given most folks behind these changes will consider backward compatibility and making things pain-less for other implementations, it should be generally safe for us to just ignore the tag type and treat the post as is, rather than to raise exception and discard the whole post.

Personally, I think fep-e232 inline quote is really interesting, and backed by most implementations like Pleroma/Misskey/Firefish except Mastodon, but that can be its own pr in the future.

### DIFF
--- a/activities/models/post.py
+++ b/activities/models/post.py
@@ -909,13 +909,11 @@ class Post(StatorModel):
                 elif tag_type in ["toot:emoji", "emoji"]:
                     emoji = Emoji.by_ap_tag(post.author.domain, tag, create=True)
                     post.emojis.add(emoji)
-                elif tag_type == "edition":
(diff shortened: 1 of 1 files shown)


andrewgodwin (approved):

andrewgodwin:
Agreed - a lot of this code was written to raise exceptions when I was developing it so I could discover compat holes, but that's less useful now.

PR #641, opened 2023-09-01
add relay support

alphatownsman:
- [x] initialize local Relay actor in post-migration 
- [x] subscribe to relay server
- [x] fanout public posts to active relay server
- [x] accept inbound relayed post (no work needed)
- [x] unsubscribe from relay server 
- [x] UI to subscribe/unsubscribe relay

## Protocol
from `users/models/relay_actor.py`:
```
    Relay services typically supports two modes: Mastodon and LitePub.

    in Mastodon mode:
    instance handshake with relay by sending Follow[1] to relay's inbox, relay won't follow back
    instance publishes its activities by sending Create/Update/Delete to relay's inbox
    relay publishes remote activities by sending Announce to instance's inbox

    in LitePub mode (also known as Pleroma mode):
    instance queries relay server's actor uri, sends Follow to relay's inbox and gets Follow back
    instance publishes its activities by sending Announce[2] to relay's inbox
    relay publishes remote activities by sending Announce to instance's inbox

    Our implementation do initial handshake in LitePub mode, then send activities in Mastodon mode;
    Most if not all modern relay implementations seem fine with this mixed mode

```
The mixed mode is chosen to minimize necessary hacks to the current code and data model

## Test

The PR has been tested with instances running the following relay implementations
- https://git.pleroma.social/pleroma/relay/
- https://github.com/yukimochi/Activity-Relay
- https://git.asonix.dog/asonix/relay
- https://github.com/noellabo/pub-relay/ - not working, blocked by [a strict check](https://github.com/noellabo/pub-relay/issues/24) I don't feel very comfortable to accommodate

## Usage

Relay can be added in admin UI (screen shot below), or run the following in `python manage.py shell` :
```
from users.models import RelayActor
RelayActor.subscribe('https://relay.infosec.exchange/actor')
RelayActor.unsubscribe('https://relay.infosec.exchange/actor')
```
be aware: large relay may take minutes or longer to respond your subscribing request, and once it does, large volume of activities will be posted to your server at once.

Screenshot 2023-09-01 at 16 49 27

### DIFF
--- a/activities/models/post_interaction.py
+++ b/activities/models/post_interaction.py
@@ -10,7 +10,6 @@ from core.ld import format_ld_date, get_str_or_id, parse_ld_date
 from core.snowflake import Snowflake
 from stator.models import State, StateField, StateGraph, StatorModel
 from users.models.identity import Identity
-from users.models.relay_actor import RelayActor
 
 
 class PostInteractionStates(StateGraph):
@@ -299,11 +298,6 @@ class PostInteraction(StatorModel):
                 "object": self.post.object_uri,
                 "to": "as:Public",
             }
-            if self.identity.is_local_relay:
-                # if boost to relay, set "to" to remote relays instead of Public
-                value["to"] = list(
-                    RelayActor.get_relays().values_list("actor_uri", flat=True)
-                )
         elif self.type == self.Types.like:
             value = {
                 "type": "Like",
@@ -321,7 +315,7 @@ class PostInteraction(StatorModel):
                 "inReplyTo": self.post.object_uri,
                 "attributedTo": self.identity.actor_uri,
             }
-        else:
+        elif self.type == self.Types.pin:
             raise ValueError("Cannot turn into AP")
         return value
 
--- a/templates/admin/identities.html
+++ b/templates/admin/identities.html
@@ -52,7 +52,6 @@
                         Remote
                         <small>{{ identity.followers_count }} local follower{{ identity.followers_count|pluralize }}</small>
                     {% endif %}
-                    {{ identity.actor_type }}
                 </td>
                 <td class="actions">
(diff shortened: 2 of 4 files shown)


activities/migrations/0001_initial.py
```diff
@@ -264,6 +263,7 @@ class Migration(migrations.Migration):
                             ("identity_edited", "Identity Edited"),
                             ("identity_deleted", "Identity Deleted"),
                             ("identity_created", "Identity Created"),
+                            ("identity_moved", "Identity Moved"),
```
andrewgodwin, line 266: I'll do this on main, it shouldn't need to be in here.

andrewgodwin (commented):
So, can you explain why you think a separate Relay actor is required, that's not SystemActor nor a standard Identity? From scanning through, it doesn't seem like it does anything that either of those couldn't be extended to do.

andrewgodwin:
The shared inbox thing we can do with system actor, but what software _requires_ that the relay actor ends in '/relay'? That seems... strange.

andrewgodwin:
What sort of hack is required for "mastodon mode"? I'm curious, as it would seem less fragile than mixing the modes as is done currently.

andrewgodwin:
I might say that, instead of doing special things with Identities, we should just have a new Relay model that does this instead? It can have a "unsent"/"sent" state to handle this follow, and it would made the admin code cleaner as well.

andrewgodwin:
Well, Takahē will accept anything posted to its inbox URLs without even needing a follow object for it - internally, it treats every inbound message the same (shared inbox or personal inbox) and works out its own fanout, so we really don't need to even know about the relay for inbound content.

Thus, that means the only thing we'd be using Identity for is the follow accept/reject logic; the message acceptance doesn't need an identity (it can use either the shared inbox URI, or we can invent one with relay in the name that goes to the same place if we want), and sending every local message out to a relay has to be special code anyway that adds a fanout for every relay in the targets calculation code.

PR #643, opened 2023-09-16
Update ld schema to support multikey and wytchspace implementations

humrochagf:
After watching the latest json-ld canonicalize exceptions the most common are coming from instances that implement those two contexts (Both are failing due to bad redirection):

- https://w3id.org/security/multikey/v1
- https://purl.org/wytchspace/ns/ap/1.0

For the first one, I'm fixing it by bringing the entire context to the schema since it's relatively small and is more commonly used.

For the second one, I just added the proper redirection inside the context since it's less common and bringing the entire context is less valuable.

### DIFF
--- a/core/ld.py
+++ b/core/ld.py
@@ -458,6 +458,46 @@ schemas = {
             }
         },
     },
+    "w3id.org/security/multikey/v1": {
+        "contentType": "application/ld+json",
+        "documentUrl": "https://w3id.org/security/multikey/v1",
+        "contextUrl": None,
+        "document": {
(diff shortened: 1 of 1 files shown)


andrewgodwin (approved):

andrewgodwin:
Lovely, thank you!

osmaa:
what benefit are we getting from building in that wytchspace AP context?

andrewgodwin:
Overcoming parsing issues before we fix them as we were discussing them on #644, I'd say!

Issue #640, opened 2023-08-28
JsonLdError in canonicalise when AP event contains LD schema not built-in


osmaa:
Repetitive error in my logs:

```
('Dereferencing a URL did not result in a valid JSON-LD object. Possible causes are an inaccessible URL perhaps due to a same-origin policy (ensure the server uses CORS if you are using client-side JavaScript), too many redirects, a non-JSON response, or more than one HTTP Link Header was provided for a remote context.',)
Type: jsonld.InvalidUrl
Code: loading remote context failed
Details: {'url': 'https://app.wafrn.net/contexts/litepub-0.1.jsonld', 'cause': JsonLdError("No schema built-in for '*/contexts/litepub-0.1.jsonld'")}
```

```
("No schema built-in for '*/contexts/litepub-0.1.jsonld'",)
Type: jsonld.LoadDocumentError
Code: loading document failed
Cause: KeyError  File "/takahe/core/ld.py", line 587, in builtin_document_loader
    return schemas[key]
           ~~~~~~~^^^^^
```

Changing the built-in schema wildcard for litepub doesn't seem like the right choice for one server placing its schemas in an untypical path. Mastodon loads remote schemas, and if we did too, that'd solve this, as the URL does resolve. However, we don't load remote schemas, and if we did, we probably should have a cache for them.

andrewgodwin:
Yeah, this was one of the tradeoffs when I made it not do remote queries for schemas; it'll need fixing eventually. I do wonder if we could just strip out unknown schemas for now though, since I doubt we use anything referred to by them.

osmaa:
okay, I'll just test not raising the error, see what happens.

AstraLuma:
Was this fixed by #644?

osmaa:
it was

PR #644, opened 2023-09-18
deal with unknown json-ld schema

osmaa:
ref #640 this implements a remote schema loader with a tiny in-memory cache for not-built-in contexts, and also returns an empty schema when one can't be loaded (or is not valid json) from the remote server. This entirely eliminated the largest source of errors in my sentry monitoring even without #643.

Introduces a new dependency with httpx-cache. A simpler version of this fix would be to just replace the `JsonLdError` in `builtin_document_loader` with a `return {}`, since nothing depends on not canonicalizing schema which hasn't been built-in.

### DIFF
--- a/core/apps.py
+++ b/core/apps.py
@@ -1,7 +1,7 @@
 from django.apps import AppConfig
 from pyld import jsonld
 
-from core.ld import caching_document_loader
+from core.ld import builtin_document_loader
 
 
 class CoreConfig(AppConfig):
@@ -9,4 +9,4 @@ class CoreConfig(AppConfig):
     name = "core"
 
     def ready(self) -> None:
-        jsonld.set_document_loader(caching_document_loader)
+        jsonld.set_document_loader(builtin_document_loader)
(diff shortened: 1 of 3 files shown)


andrewgodwin:
Yeah, to be honest I'd rather go with the `return {}`, since ActivityStreams specifically says that messages should adhere to an exact main context and that you don't need a full JSON-LD parser to decode them; the less network requests we can do for things we're not going to use, the better.

osmaa:
yeah, I haven't noticed any benefit to pulling these remote schemas - quite the contrary, many AP messages apparently have context references to 404 URLs.

andrewgodwin:
Yes, I suspect that some of them are just used as flags for internal parsers or similar. Want to flip this PR to just return empty context, or open a new one for that?

osmaa:
context loader and cache reverted, blank schema added to the builtins, returning that instead of raising exception

andrewgodwin (approved):

andrewgodwin:
Lovely, thanks for that!

PR #645, opened 2023-09-18
separate out timeouts from other remote server issues

osmaa:
Another relatively large group of errors in my sentry are remote fetches for webfinger or pinned posts from slow servers. Raising a TryAgain instead of a fatal error fixes majority of these - however, one can't ensure a canonical actor URI is known until webfinger has been successful.

### DIFF
--- a/tests/users/models/test_identity.py
+++ b/tests/users/models/test_identity.py
@@ -168,6 +168,14 @@ def test_fetch_actor(httpx_mock, config_system):
             "url": "https://example.com/test-actor/view/",
         },
     )
+    httpx_mock.add_response(
+        url="https://example.com/test-actor/collections/featured/",
+        json={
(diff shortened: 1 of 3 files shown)


andrewgodwin (approved):
Makes sense to me!

osmaa:
mock featured posts added to the test, though the test coverage isn't quite as deep as I would have liked. Also added handling of TryAgainLater to users/views/activitypub.py. Did not move the exception from stator to core, that probably would be advisable?

andrewgodwin (approved):

andrewgodwin:
I'm trying to keep Stator separate almost like a library in case I spin it out, so I do want to keep TryAgainLater in there as it's part of its direct runtime behaviour.

Thanks for this!

PR #646, opened 2023-10-23
use logging instead of sentry.capture_*

osmaa:
Replace sentry integration's capture_message and capture_exception with Python logging and Sentry's built-in integration to it. This has the following benefits:

- improves logs when Sentry is not used
- improves contextuality, as logging.info() become Sentry breadcrumbs
- reduces footprint of the Sentry-specific code (though scope setup/clear remains)

### DIFF
--- a/activities/models/post.py
+++ b/activities/models/post.py
@@ -1,5 +1,6 @@
 import datetime
 import json
+import logging
 import mimetypes
 import ssl
 from collections.abc import Iterable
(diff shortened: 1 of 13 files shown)


andrewgodwin (approved):

andrewgodwin:
Thanks - seems sensible enough!

kfdm:
Sorry for the drive by comment, but I'm curious about something.

Typically if you use something like `logging.info` then it logs everything to the root logger. While not specifically an issue, often it seems better to get an instance of a logger so that the logs also show where in the file things are.

```
import logging
logger = logging.getLogger(__name__)
...
logger.info('log like usual')
```

andrewgodwin:
Yes, that's true, but Sentry's integration has a bit more locality due to the breadcrumb behaviour, so while this PR isn't perfect, it's still an improvement over the previous code!

Issue #642, opened 2023-09-03
IntegrityError at /api/v2/search when searching remote handles


Ultimator14:
I'm using Tusky as client app for my own takahe server. In the takahe configuration I enabled `ERROR_EMAILS`. 
Whenever I search for a (remote) handle the first time e.g. `@takahe@jointakahe.org` using the Tusky search, I get two mails informing me about a server error.

```
Internal Server Error: /api/v2/search

IntegrityError at /api/v2/search
doppelter Schlüsselwert verletzt Unique-Constraint »users_identity_actor_uri_key«
DETAIL:  Schlüssel »(actor_uri)=(https://jointakahe.takahe.social/@takahe@jointakahe.org/)« existiert bereits.
```
(full message see below)

This happens independently of the remote server software. I noticed this for mastodon, takahe, kbin and also some others. This only happens if the user is unknown to the server, so the very first time I do the search. Afterwards, the error does no longer occur but  only the pinned posts are shown in the Tusky app and follows are also not displayed correctly (see screenshot).

Log in error mail (The log leaks some usernames and passwords, I redacted them manually):
 ```
 Internal Server Error: /api/v2/search

IntegrityError at /api/v2/search
doppelter Schlüsselwert verletzt Unique-Constraint »users_identity_actor_uri_key«
DETAIL:  Schlüssel »(actor_uri)=(https://jointakahe.takahe.social/@takahe@jointakahe.org/)« existiert bereits.

Request Method: GET
Request URL: https://social.pygos.space/api/v2/search?q=%40takahe%40jointakahe.org&type=statuses&resolve=true&limit=20&offset=0&following=false
Django Version: 4.2.4
Python Executable: /usr/local/bin/python3
Python Version: 3.11.5
Python Path: ['/takahe', '/usr/local/bin', '/usr/local/lib/python311.zip', '/usr/local/lib/python3.11', '/usr/local/lib/python3.11/lib-dynload', '/usr/local/lib/python3.11/site-packages']
Server time: Sun, 03 Sep 2023 20:34:01 +0000
Installed Applications:
['django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django.contrib.postgres',
 'corsheaders',
 'django_htmx',
 'hatchway',
 'core',
 'activities',
 'api',
 'mediaproxy',
 'stator',
 'users']
Installed Middleware:
[...]

PR #647, opened 2023-10-23
Refactoring inbox processing to smaller tasks

osmaa:
(I sent a PR with the first commit separetely, but this builds on top and uses logging for providing diagnostics)

1. Check HTTP signature ahead of LD signature, and only strip the LD sig away if it doesn't verify.

This fixes random rejects of incoming messages, because LD Signatures just don't always check correctly due to mysterious differences in the JSON->RDF translation

2. Do not fetch_actor in the Inbox POST path, because sometimes that's just really, really slow and can DoS the web server

This also means that we can't verify the signature of the first ActivityPub message we see from a new actor - because the key isn't yet known. Stator will fetch it in the second stage of processing and it'll be available for further messages.

Without this change, my server was sporadically bogged down by storms of messages from *blocked* domains, because fetch_actor was called before checking for domain blocks.

3. Fetch pinned posts in a decoupled internal message instead of fetch_actor

also because it slowed fetch_actor down a lot, and frequently caused database integrity errors as multiple threads were trying to insert the same data

### DIFF
--- a/activities/models/post.py
+++ b/activities/models/post.py
@@ -884,7 +884,7 @@ class Post(StatorModel):
                 except IntegrityError:
                     # despite previous checks, a parallel thread managed
                     # to create the same object already
-                    post = cls.by_object_uri(object_uri=data["id"])
+                    raise TryAgainLater()
             else:
                 raise cls.DoesNotExist(f"No post with ID {data['id']}", data)
         if update or created:
@@ -1014,7 +1014,7 @@ class Post(StatorModel):
                     response = SystemActor().signed_request(
                         method="get", uri=object_uri
                     )
-                except (httpx.HTTPError, ssl.SSLCertVerificationError):
+                except (httpx.HTTPError, ssl.SSLCertVerificationError, ValueError):
                     raise cls.DoesNotExist(f"Could not fetch {object_uri}")
                 if response.status_code in [404, 410]:
                     raise cls.DoesNotExist(f"No post at {object_uri}")
--- a/activities/models/post.py
+++ b/activities/models/post.py
@@ -1072,7 +1072,7 @@ class Post(StatorModel):
             if data["actor"] != data["object"]["attributedTo"]:
                 raise ValueError("Create actor does not match its Post object", data)
             # Create it, stator will fan it out locally
-            cls.by_ap(data["object"], create=True, update=True)
+            cls.by_ap(data["object"], create=True, update=True, fetch_author=True)
 
     @classmethod
     def handle_update_ap(cls, data):
--- a/activities/models/post.py
+++ b/activities/models/post.py
@@ -1,5 +1,6 @@
 import datetime
 import json
+import logging
 import mimetypes
 import ssl
 from collections.abc import Iterable
@@ -27,7 +28,7 @@ from activities.models.post_types import (
     PostTypeDataEncoder,
     QuestionData,
 )
-from core.exceptions import ActivityPubFormatError, capture_message
+from core.exceptions import ActivityPubFormatError
 from core.html import ContentRenderer, FediverseHtmlParser
 from core.ld import (
     canonicalise,
@@ -896,7 +897,7 @@ class Post(StatorModel):
                 # don't have content, but this shouldn't be a total failure
                 post.content = get_value_or_map(data, "content", "contentMap")
             except ActivityPubFormatError as err:
-                capture_message(f"{err} on {post.url}")
+                logging.warning(f"{err} on {post.url}")
                 post.content = None
             # Document types have names, not summaries
             post.summary = data.get("summary") or data.get("name")
@@ -992,7 +993,7 @@ class Post(StatorModel):
                     try:
                         cls.ensure_object_uri(post.in_reply_to, reason=post.object_uri)
                     except ValueError:
-                        capture_message(
+                        logging.warning(
                             f"Cannot fetch ancestor of Post={post.pk}, ancestor_uri={post.in_reply_to}"
                         )
                 else:
--- a/activities/services/post.py
+++ b/activities/services/post.py
@@ -1,3 +1,5 @@
+import logging
+
 from activities.models import (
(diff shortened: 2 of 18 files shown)


andrewgodwin:
Seems like a decent approach - could you rebase onto `main` now I took in the other PR so the diff is easier to read?

osmaa:
Rebased. Probably fixes #642

osmaa:
Sorry for the test spam there. I don't have a working test environment locally..

The test fail in signatures confuses me. The code should still raise VerificationError on an invalid signature - what has changed there is that:

1. if the signature is valid, it's not stripped out of the document (original code popped the sig out of document passed by reference)
2. the VerificationError is caught in activitypub.Inbox.post and dealt with not by returning HttpResponseUnauthorized, but simply by stripping out the invalid signature, which is encountered frequently enough in the wild for reasons I can't fully track down (apparently are related to canonicalization and/or the normalized hash, though)

Reverting the little changes I made to core/signatures.py anyway..

osmaa:
Can't see why the signature test continues to fail.

osmaa:
The reason I started looking into the signatures at all was that I was seeing daily failures in LD signature verification from messages I would see Mastodon servers having accepted. I spent a lot of time trying to decipher why -- the only thing I could find out was that sometimes sample JSON would show up with different RDF encoding to what pyld would produce, but the whole process is very opaque and I could get no explanation to that at all.

But in the process I also found out that Mastodon doesn't recommend implementing LD Signatures at all, because their currently implemented version is obsolete, and that Mastodon only refuses to forward LD sigs which don't check out, but doesn't refuse the message itself.

Which is what I did here as well:
- check HTTP sig first
- don't reject messages for LD Signature fail
- DO keep the LD sig on documents where the verification succeeds (this is still the majority of LD signed messages)

I only noticed now that while the signature check still works for *correct* LD Sigs, the test case for a manipulated message doesn't raise an exception as the test expects. Except... the implementation still raises an exception in the same cases as before, and running the code in the wild, I see messages in my logs about LD signatures being stripped of messages when they don't check out.

At first blush, it would seem to me the test is broken. Except.. I didn't change the test either, and it must have worked before.

I know, I should have a working local test environment and having run the tests before making any changes - or reproduce them without this patch. I haven't gotten around to setting up an env with postgres available for the tests, though :(

osmaa:
It is the test that is busted. The test document:

```
{
        "id": "https://example.com/test-create",
        "type": "Create",
        "actor": "https://example.com/test-actor",
        "object": {"id": "https://example.com/test-object", "type": "Note"},
        "signature": {
            "@context": "https://w3id.org/identity/v1",
            "creator": "https://example.com/test-actor#test-key",
            "created": "2022-11-12T21:41:47Z",
            "signatureValue": "nTHfkHqG4hegfnjpHucXtXDLDaIKi2Duk+NeCzqTtkjf4NneXsofbZY2tGew4uAooEe1UeM23PIyjWYnR16KwcD4YY8nMj8L3xY2czwQPScMM9n+KhSHzkWfX+iI4FWKbjpPI8M53EtTRJU+1qEjjmGUx03Ip0vfvT5821etIgvY4wLNhg3y7R8fevnNux+BeytcEV6gM4awJJ6RK0xrWGLyTgDNon5V5aNUjwcV/UVPy9UAQi1KYWtA74/F0Y4oPzL5CTudPpyiViyVHZQaal4r+ExzgSvGztqKxQeT1ya6gLXxbm1YQ+8UiGVSS8zoGhMFDEZWVsRPv7e0jm5wfA==",
            "type": "RsaSignature2017",
        },
    }
```

has no `@context` for the top level, so jsonld.normalize will turn it into an empty string instead of the expected n-quads. The normalized_hash for it will be the same before and after the "evil" modification.. Why did it work before and what else has broken it? There were some changes to the builtin_document_loader contexts before, but I can't see how that could have changed anything here: the test doc just has no context to begin with.

andrewgodwin (approved):

andrewgodwin:
Ahh, nice catch. I had no idea Mastodon had deprecated LD sigs - I wonder what the situation is like for other servers.

osmaa:
My understanding is that Mastodon servers with AUTHORIZED_FETCH enabled won't produce LD Sigs at all - because the idea of the feature is to require subscribers to authenticate with the source server to see the content, and removing LD Sigs creates plausible deniability over the content.
https://docs.joinmastodon.org/admin/config/#authorized_fetch

I had a quick peek at Pixelfed and Firefish source. Pixelfed doesn't appear to contain LD Signature support at all, and Firefish apparently only attaches LD Sigs to messages in the relay service. Makes sense, actually.

andrewgodwin:
Yeah, that does make sense. The ones I'd be mostly thinking that did this would be gotosocial and pleroma; they tend to be the most prevalent in terms of content I've seen.

osmaa:
I found only HTTP signature implementation in both gotosocial and pleroma, no sign of LD sig in either. And for whatever it's worth, now that (valid) LD sigs are retained on the inbox messages, scanning that table (sorry for the pretty nasty SQL here), I only see LD sigs on Mastodon and Hometown servers (the latter being a fork of Mastodon). The invalid LD sigs are logged, but that being a text file I didn't bother checking all of them for software type. On a spot check, I see a several of those log messages caused by the same Mastodon servers I see valid sigs from -- that being an indication that LD Sigs aren't a stable thing due to something in the normalization/RDF translation process.

```
select t1.domain,software,signatures from 
(select split_part(actor, '/',3) domain,count(signature) signatures from 
  (select id, message->>'actor' actor , message->'signature'->>'type' signature from users_inboxmessage ) a 
 where signature='RsaSignature2017' group by 1) t1 
join 
(select domain,nodeinfo->'software'->>'name' software from users_domain) t2 
on (t1.domain=t2.domain) 
order by signatures desc;
```

andrewgodwin:
Great, thanks - sounds like there's really not much point continuing to verify or implement them, really...

PR #648, opened 2023-10-28
focalpoints are floats between -1..1, not int

osmaa:
https://docs.joinmastodon.org/api/guidelines/#focal-points

(another thing caught via a sentry alert)


andrewgodwin:
Does this not need a database migration?

osmaa:
it does. I wasn't sure what to do with the autogenerated file

osmaa:
adding the test case for this i noticed that attachment.to_ap() wasn't producing the (under-specced) Mastodon focalPoint tuple, so that's now added, too.

What I haven't tested is whether defining a focal point through a client app actually goes through end-to-end. I suppose it should, now..

activities/models/post_attachment.py
```diff
@@ -122,6 +122,11 @@ def to_ap(self):
             "mediaType": self.mimetype,
             "blurhash": self.blurhash,
         }
+        if self.is_image() and self.focal_x and self.focal_y:
+            ap["type"] = "Image"
+            ap["mediaType"] = self.mimetype
```
andrewgodwin, line 127: This seems redundant?

osmaa: mediatype, you're right. for type = "Image", not sure. Mastodon sets and sends focalpoint on type = Image attachments, and it doesn't seem right to try to set that on type = Document attachments instead. This is one of those cases where end-to-end interop testing is not simple - it turns out none of the clients I use even support defining a focal point, so I can't send a test message..

andrewgodwin (approved):

activities/models/post_attachment.py
andrewgodwin: Yeah, it was just the double mediaType assignment I was calling out!

andrewgodwin:
Thanks!

PR #654, opened 2023-11-15
Do not canonicalise non json content in the search endpoint

humrochagf:
While testing https://phanpy.social/ web client I noticed it calls the search API with the links found in the posts to do some clever discovery magic.

But by doing so it makes the service to search for things that returns with 200 but are not json content, so we get some serialization crashes.

This checks for the response `Content-Type` before trying to do json parsing and canonicalization.


andrewgodwin:
:+1:

PR #657, opened 2023-11-16
Module-specific loggers and minor reformatting

osmaa:
Addressing https://github.com/jointakahe/takahe/pull/646#issuecomment-1776286087 by initializing a logger in each module, reformatting a few messages, and a little bit of additional sentry context work.


andrewgodwin (approved):

andrewgodwin:
Thanks! I'll verify that the Stator logs still work locally (#646 disabled them it turns out, but I don't think this will)

PR #659, opened 2023-11-17
allow unicode characters in hashtag

alphatownsman:


andrewgodwin:
I'd like some tests with this that verify that it does, indeed, allow unicode characters that can be considered words (and does not allow, say, whitespace classes). The existing tests are in `tests/core/test_html.py` - you can add a few cases under the existing hashtag test in `test_parser`.

andrewgodwin (approved):

andrewgodwin:
That looks good - thanks!

PR #661, opened 2023-11-19
fix fetching post from another takahe by searching its url

alphatownsman:
This fixes the issue that searching a remote takahe post url not yet on local server will fail.


andrewgodwin (approved):

andrewgodwin:
Thanks!

PR #662, opened 2023-11-19
Improve search api json parsing

humrochagf:
The `Content-Type` http header not only can have the actual type of the payload but also can return the charset and other optional parameters like defined in [HTTP/1.1 RFC](https://www.rfc-editor.org/rfc/rfc7231#section-3.1.1.5).

So Content-Type like the following were getting skipped by our checks:

```text
Content-Type: text/html; charset=utf-8
```

To make it more robust I took in consideration the charset to parse the json content from the request, and also added some tests to cover this endpoint.

### DIFF
--- a/activities/services/search.py
+++ b/activities/services/search.py
@@ -1,3 +1,5 @@
+import json
+
 import httpx
 
 from activities.models import Hashtag, Post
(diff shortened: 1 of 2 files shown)


andrewgodwin (approved):

andrewgodwin:
Amazing, thank you!

PR #663, opened 2023-11-20
Extract json parser to core and use in fetch_actor

humrochagf:
Following with the improvement to the search #662 there's a second fetch that is made by the fetch_actor during the search request to the api, so when adding characters that may break in the wrong encoding to the payload we can see the need to make the same fix to the fetch_actor method.

I then extracted the function to core module so it can be reused.

Obs.: We may need to reuse it in other places where we fetch json or json-like content, but I'll keep this PR sort by only covering the path impacted by the test.

### DIFF
--- a/activities/services/search.py
+++ b/activities/services/search.py
@@ -1,8 +1,7 @@
-import json
-
 import httpx
 
 from activities.models import Hashtag, Post
+from core.json import json_from_response
 from core.ld import canonicalise
 from users.models import Domain, Identity, IdentityStates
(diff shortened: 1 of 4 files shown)


andrewgodwin (approved):

andrewgodwin:
Seems sensible - it might be worth me doing a similar breakout for the new separate ActivityPub library I'm working on that we'll eventually move Takahē to (https://github.com/jointakahe/taktivitypub), since this seems to be a common need.

PR #666, opened 2023-11-26
do not retry unmute if mute never expires

alphatownsman:
it seems currently the mute state stuck in `awaiting_expiry` and loop every hour even if it does not have to.

let me know if I should add migration like
```
update users_block set state = 'sent' where mute=true and expires is null;
```

### DIFF
--- a/users/models/block.py
+++ b/users/models/block.py
@@ -195,8 +195,7 @@ class Block(StatorModel):
             raise ValueError("You cannot mute from a remote Identity")
         block = cls.maybe_get(source=source, target=target, mute=True)
         if block is not None:
-            if not block.active:
-                block.state = BlockStates.new  # type:ignore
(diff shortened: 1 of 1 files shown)


andrewgodwin:
The bug here appears to be that there is a `sent` state at all - I intended for all mutes to live in awaiting_expiry even if they aren't expirable when I wrote this. This patch isn't sufficient by itself to get to the end effect of not checking these states; I'd instead want to see:

* The state renamed to something like `permanent`
* `create_local_mute` updated to correctly move a mute out of that state back into `new` if it has its duration updated at all

andrewgodwin (approved):

andrewgodwin:
Yeah, with the `create_local_mute` patch I think that's good for now - and you're right, the sharing of the model with Block does complicate things somewhat.

PR #667, opened 2023-11-26
do not make local identity outdated

alphatownsman:
it seems currently local identity flip back to `outdated` every 15 days unnecessarily, this commit fixes that. 

a better solution might be having a dedicated `local_updated` state which never retry, but that requires much more change and I'll save it for another day.


andrewgodwin (approved):

andrewgodwin:
Nice catch!

I agree with the other state, but yes, definitely a bigger patch.

PR #669, opened 2023-11-26
make max_media_attachments configurable by admin

ishigoya:


users/views/admin/settings.py
```diff
@@ -38,6 +38,10 @@ class BasicSettings(AdminSettingsPage):
             "title": "Maximum Post Length",
             "help_text": "The maximum number of characters allowed per post",
         },
+        "max_media_attachments": {
+            "title": "Maximum Media Attachments",
+            "help_text": "The maximum number of media attachments allowed per post.\nA higher value may not be supported by some servers or clients.",
```
andrewgodwin, line 43: Can we change this to "A setting that is not 4 may be unsupported by clients"?

andrewgodwin (approved):

andrewgodwin:
Thanks!

PR #674, opened 2023-12-01
Add --log-config cmd line parameter to stator

osmaa:
ref https://github.com/jointakahe/takahe/commit/ab3648e05d4d532800d9f4a6d1ecfbca28f8e3d8#commitcomment-132861912

With this, it's possible to use the same logging config ini file for both the web and stator processes. Sample logging config included.

### DIFF
--- a/docker/logging_config.ini
+++ b/docker/logging_config.ini
@@ -1,0 +1,57 @@
+[handlers]
+keys = out, error
+
+[handler_out]
+class = StreamHandler
+formatter = generic
+args = (sys.stdout,)
+
+[handler_error]
+class = StreamHandler
+formatter = generic
+args = (sys.stderr,)
+
+[formatters]
+keys = generic
+
+[formatter_generic]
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Well, I more mean that it's odd to check a default in but not have it used; you could either make it an example one not in the docker directory, place it next to the code and have it as a relative-path default, or include it in the documentation.

osmaa:
Personally, I mount the config into the running container and then refer to it on the starting command line, because that allows me to tweak the config without rebuilding the image. That probably isn't what everyone wants to do, though..

andrewgodwin:
Yeah, my deployment infra assumes that a redeploy means a new image. It's mostly a matter of personal taste I suspect.

AstraLuma:
So what changes do you want to see, @andrewgodwin ?

andrewgodwin:
Well, if there's going to be a default config added to the docker image, I think the image should be set up to use that via environment variable or something.

PR #676, opened 2023-12-06
when remote follows local, hold off sending Accept if remote identity not fully fetched

alphatownsman:
When a new remote identity follows a local identity (manual approval = false), `Follow.by_ap` will create a `Follow` and a bare bone `Identity`, so these will be triggered:

 - `IdentityStatus.handle_outdated`
 - `FollowStates.handle_unrequested` 
 - `FollowStates.handle_accepting` 

there are rare cases `FollowStates.handle_accepting` get executed before `IdentityStatus.handle_outdated`, which will raise [exception](https://neodb.sentry.io/share/issue/689fe9c3ae674335bd9a4b137400e976/) bc missing `inbox_uri`

This PR should fix this.

### DIFF
--- a/users/models/follow.py
+++ b/users/models/follow.py
@@ -17,7 +17,7 @@ logger = logging.getLogger(__name__)
 class FollowStates(StateGraph):
     unrequested = State(try_interval=600)
     pending_approval = State(externally_progressed=True)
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Thanks!

PR #678, opened 2023-12-12
improve identity deletion

alphatownsman:
- disable all current login tokens
 - ignore new mention
 - ignore new follow
 - delete all data
 - fanout delete posts, interactions
 - fanout unfollow both directions
 - fanout nullified identity, then fanout delete identity
 - keep nullified identity for 7 days then delete forever

### DIFF
--- a/users/models/identity.py
+++ b/users/models/identity.py
@@ -120,12 +120,47 @@ class IdentityStates(StateGraph):
 
     @classmethod
     def handle_deleted(cls, instance: "Identity"):
-        from activities.models import FanOut
+        from activities.models import (
+            FanOut,
+            Post,
+            PostInteraction,
+            PostInteractionStates,
+            PostStates,
+            TimelineEvent,
+        )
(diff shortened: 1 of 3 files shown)


users/models/identity.py
```diff
@@ -120,12 +120,47 @@ def handle_moved(cls, instance: "Identity"):
...
+            PostInteractionStates,
+            PostStates,
+            TimelineEvent,
+        )
+        from users.models import Bookmark, Follow, FollowStates, HashtagFollow, Report
 
         if not instance.local:
             return cls.updated
 
+        # Delete local data including logins
```
andrewgodwin, line 136: This says "including logins" but it does not seem to (and I would not want it to as that's on the User level) - is this an outdated comment?

andrewgodwin:
Great - get the lint failures passing (just run `pip install pre-commit; pre-commit run -a` locally if you've not installed the hooks yet) and I think we can merge this in.

andrewgodwin (approved):

andrewgodwin:
Thanks!

PR #679, opened 2023-12-12
Takahē HTTP Client

### LINKED ISSUE #402: Guard against fetching local IP ranges
Anything that resolves to local IP ranges should not be allowed for actor fetching.

AstraLuma:
Adds `core.httpy`, which is a wrapper around HTTPX, to provide the features:

* Block requests to some IP ranges (#402)
* Cleaner signature handling (just pass the HTTP client the actor you're using)

Fixes #402

### DIFF
--- a/core/httpy.py
+++ b/core/httpy.py
@@ -1,0 +1,62 @@
+"""
+Wrapper around HTTPX that provides some fedi-specific features.
+
+The API is identical to httpx, but some features has been added:
+
+* Fedi-compatible HTTP signatures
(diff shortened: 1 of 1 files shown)


AstraLuma:
I can also make HTTP/2 default, if there aren't any concerns about that. (HTTPX does some vague self-FUD in the docs.)

AstraLuma:
@andrewgodwin before I go and Convert All The Things, can I get a design review? I don't want to have to change 50 calls later.

PR #680, opened 2023-12-13
basic protection against invalid domain names

alphatownsman:
current behavior is very insecure, this patch solves part of problem but not all 

e.g. https://mysite.com/@abc@whatever/posts/123/ will create garbage data in database, which is a bit insecure, but the data is at least marked as `connection_issue` after some stator cycle. however, if attacker uses https://mysite.com/@abc@whatever@mysite.com/posts/123/ , the domain `whatever@mysite.com` in db will look valid (state=updated and local=False), which might be used to construct further attacks. this patch solves the latter issue by validate domain before saving to db. this patch also prevent `localhost` from used

a future patch IMHO should be implemented to protect against more cases.

### DIFF
--- a/tests/users/models/test_domain.py
+++ b/tests/users/models/test_domain.py
@@ -3,6 +3,36 @@ import pytest
 from users.models import Domain
 
 
+def test_valid_domain():
+    """
+    Tests that a valid domain is valid
+    """
+
+    assert Domain.is_valid_domain("example.com")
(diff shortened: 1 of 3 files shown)


andrewgodwin:
Hmm, I'm not entirely sure about not allowing IP addresses as I think they're technically valid, but I agree we probably don't actually want them in reality.

andrewgodwin (approved):

PR #683, opened 2023-12-26
Fix accept/reject remote follow request

alphatownsman:
When accept or reject a remote follow request from Pleroma/Firefish/Misskey/etc, the `id` in the message should match the local actor, at least for the host name part.

The issue was brought up previously and correctly pinpointed in #323. However, the last commit in #323 seemed incorrect. This PR intend to fix that for good.

### DIFF
--- a/users/models/follow.py
+++ b/users/models/follow.py
@@ -282,7 +282,7 @@ class Follow(StatorModel):
         """
         return {
             "type": "Accept",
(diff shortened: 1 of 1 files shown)


andrewgodwin (approved):

andrewgodwin:
Thanks - that one's on me!

PR #684, opened 2023-12-27
post update/delete also fanout to those who liked/boosted it but not following the author

alphatownsman:
this is useful as Mastodon will send notification if a post you liked gets edited.

### DIFF
--- a/activities/models/post.py
+++ b/activities/models/post.py
@@ -765,6 +765,8 @@ class Post(StatorModel):
         targets = set()
         for mention in self.mentions.all():
             targets.add(mention)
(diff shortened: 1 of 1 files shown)


andrewgodwin (approved):

andrewgodwin (changes requested):
I thought this would be OK at first, but I think it's got a slight security hole, where if someone changes a post's visibility this still leaks edits to it outside of that. Could you wrap it in a test that ensures the visibility is public or unlisted?

andrewgodwin (approved):

andrewgodwin:
Thanks!

PR #685, opened 2024-01-02
Workaround: 'as:Public' not recognized by many implementations

alphatownsman:
Many implementations, e.g. Pleroma and many relays, do not recognize `as:Public`, but only know `https://www.w3.org/ns/activitystreams#Public` in to/cc for public posts. This causes public posts from Takahe being rejected or ignored.

This patches outbound json to use full uri instead of as:Public. This is more of hack than a proper implementation. Suggestions and improvements are welcomed.

### DIFF
--- a/core/ld.py
+++ b/core/ld.py
@@ -647,7 +647,9 @@ def builtin_document_loader(url: str, options={}):
             return schemas["unknown"]
 
 
-def canonicalise(json_data: dict, include_security: bool = False) -> dict:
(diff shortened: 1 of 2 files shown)


andrewgodwin:
The correct way to do this is to instead change the context that the JSON-LD canonicaliser uses, to remove the "as:Public" short identifer - that way the JSON-LD library will leave it expanded.

PR #686, opened 2024-01-02
Relay support

alphatownsman:
- Admin UI to subscribe/unsubscribe from relay server (use inbox mode)
- Fan out public posts to active relay servers

### DIFF
--- a/users/models/relay.py
+++ b/users/models/relay.py
@@ -44,8 +44,8 @@ class RelayStates(StateGraph):
                 uri=instance.inbox_uri,
                 body=instance.to_follow_ap(),
             )
-        except:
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Well the problem is that we've now reached my cutoff time for keeping working on the project if nobody turns up to help full time, so there is a chance this may never get merged. It'll definitely have to wait at least until there's another maintainer who can help review it, sadly :(

AstraLuma:
Given the vibe of relay discussion, this seems like a surprisingly small change? Does this cover inbound, outbound, and handshaking?

AstraLuma:
Ok, thanks. I'm going to take some time to go over this carefully because the state machines smell of fiddly and precision.

PR #687, opened 2024-01-02
add missing migration

alphatownsman:


andrewgodwin:
This looks like it's from a Django upgrade or similar and is not actually affecting the underlying DB column - it should be instead backported to be in the first migration the field appears in, once the Django version in the dockerfile is upgraded.

andrewgodwin:
Ah, gotcha!

PR #692, opened 2024-01-19
return 410 Gone for deleted identity

alphatownsman:
instead of physically deleting a deleted identity, a more common and better practice is to return 410 Gone with type = Tombstone

### DIFF
--- a/users/views/identity.py
+++ b/users/views/identity.py
@@ -1,4 +1,5 @@
 import string
+from datetime import timezone
 
 from django import forms
 from django.contrib.auth.decorators import login_required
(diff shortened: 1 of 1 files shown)


tobru:
@alphatownsman Has this worked? I want to shut down a Takahe instance and I feel that would be a safe way to communicate to other instances that the identity is gone, rather than just deleting the instance.

users/views/identity.py
```diff
@@ -64,10 +66,13 @@ def serve_actor(self, identity):
         # If this not a local actor, redirect to their canonical URI
         if not identity.local:
             return redirect(identity.actor_uri)
-        return JsonResponse(
+        r = JsonResponse(
             canonicalise(identity.to_ap(), include_security=True),
             content_type="application/activity+json",
         )
+        if identity.deleted and tz.now() - identity.deleted > tz.timedelta(days=3):
```
AstraLuma, line 73: Can you not inline this formula? It's not complicated, but it is subtle and it took me a moment to figure out what it was doing and why.

AstraLuma (changes requested):

PR #696, opened 2024-01-30
Ensure add_stator_indexes only run for direct descendants of StatorModel

lullis:


AstraLuma:
Does this have a bug associated with it? What's the behavior you're trying to prevent?

lullis:
The issue is that these indices only make sense on models that derive directly from StatorModel. If you create any other model that itself is already derived from StatorModel (e.g,  Identity) then the system will fail the check due to issues with the indices that are not related to the table you are creating.

AstraLuma:
Ah, ok, yeah, if you have a concrete model inheriting from a concrete model ([Multi Table Inheritance](https://docs.djangoproject.com/en/5.0/topics/db/models/#multi-table-inheritance)), this code will fail because the indexed columns only exist on the parent table.

This largely looks good, but I'm going to double check some things.

stator/models.py
```diff
@@ -44,8 +44,11 @@ def add_stator_indexes(sender, **kwargs):
     Inject Indexes used by StatorModel in to any subclasses. This sidesteps the
     current Django inability to inherit indexes when the Model subclass defines
     its own indexes.
+
+    We should only add this for models that direct descendants of
+    StatorModel, otherwise we will see system check errors.
     """
-    if issubclass(sender, StatorModel):
+    if sender.__base__ is StatorModel:
```
AstraLuma, line 51: I'm not completely thrilled with this--`__base__` is semi-documented, and points to the first parent class.

But I'm not sure what multiple inheritance would do in the context of Django models.

PR #697, opened 2024-02-06
do not deliver follower-only posts to hashtag followers

alphatownsman:


AstraLuma:
This is a pretty serious leak. Unfortunately, the startup I was part of took all of my energy.

AstraLuma (commented):
I think this is right, but I want to double check:

Should unlisted posts be visible in hashtag indexes?

AstraLuma:
I'm not sure I have a personal opinion, but I don't have an intuition for what general users would want.

AstraLuma:
I talked to someone, I think that it should be changed to not delivering unlisted posts to hashtag feeds.

Especially with relays happening as soon as I can wrap my head around it.

andrewgodwin:
I would also err on the side of "don't put unlisted posts in a semi-public feed"

PR #700, opened 2024-02-19
fix potential vulnerability when fetching remote json data

alphatownsman:
when fetching remote actor or status, content type must be one of:
- "application/ld+json"
- "application/activity+json"

I did not apply this check to:
- nodeinfo
- web finger

because I don't think we have to, but open to be convinced otherwise


AstraLuma:
I think this is partially redundant with #679

PR #701, opened 2024-02-24
fix fetching Article from WriteFreely/Plume/etc

alphatownsman:


activities/models/post_types.py
```diff
@@ -110,12 +110,12 @@ class PostTypeData(BaseModel):
 
 
 class PostTypeDataEncoder(json.JSONEncoder):
-    def default(self, obj):
```
AstraLuma, line 113: why this change?

AstraLuma (approved):

PR #704, opened 2024-03-14
Allow anonymous access to reading public statuses

### LINKED ISSUE #702: Anonymous API access to statuses
The following API endpoints seems to be public by default on Mastodon, but requires the read:statuses scope on Takahe:

```
    path("v1/accounts/<id>/statuses", accounts.account_statuses),
    path("v1/statuses/<id>/source", statuses.status_source),
    path(
        "v1/statuses/<id>",
        methods(
            get=statuses.status,
        ),
    ),
    path("v1/statuses/<id>/context", statuses.status_context),
```

It would be nice if it was possible to configure Takahe to allow public access to these. 

I believe all information available with these API's is also available on the public time line (if enabled), so one solution would be to implicitly enable read:statuses for anonymous access in api/decorators.py if public time line is enabled.

pmakholm:
Fixes #702 

This allows anonymous access to API methods marked with `@scope_required("read:statuses")` if Public Timeline is enabled in the configuration.

### DIFF
--- a/api/decorators.py
+++ b/api/decorators.py
@@ -2,7 +2,7 @@ from collections.abc import Callable
 from functools import wraps
 
 from django.http import JsonResponse
-
+from core.models import Config
 
 def identity_required(function):
     """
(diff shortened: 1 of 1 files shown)


AstraLuma (approved):

AstraLuma:
To work through the design process:

This is the only place that uses `has_scope()`, but not the only place that uses the token.

The alternative I can think of is introducing a Faux Token for when the user doesn't have one. But I'm worried that'll break other things.

How easy is it to add a test for this situation?

PR #706, opened 2024-04-13
fix edge case when parsing remote json

alphatownsman:
fix exception when content type	in header is sth like `application/activity+json;`


AstraLuma (approved):
looks right

AstraLuma:
Can you add a test for this?

Issue #712, opened 2024-05-21
Current project status


ivanov17:
Hello @andrewgodwin, I found [a post on your blog about the possible sunset of the project](https://www.aeracode.org/2023/11/06/life-critical-side-projects/). Please update the readme file and the project website to reflect the current status of the project. I don't think many people will be looking for this information on your blog.

I'm very interested in this project, but unfortunately I'm not a professional programmer and can't help with it. No other ActivityPub server seems to be able to handle multiple domains. So, there is no alternative to your project. It will be sad if you close this project forever.

This is just an opinion, but maybe splitting your current project into separate parts will help get more attention from other developers. At the moment, you provide a ready-made solution and receive wishes from users. Users may share some ideas with you, but will not maintain the project. I believe developers are more interested in libraries, modules, etc., especially in the Django world.

If you're not interested in fiddling with the web interface, you can throw it away, but I think the core parts should be kept. After that you can also pass this work to community projects such as Jazzband. I believe that this is better than completely closing the project.

Thank you for your work, I believe that you truly make the world a better place. I think it would be right to start a discussion here.

AstraLuma:
Hello Ivanov! Thank you very much for your interest in the project. I am Jamie and I have taken ownership from Andrew.

I know I have been somewhat absent the last six months. My work was very stressful and did not leave room in my life for open source work of any kind. Thankfully, that's over with, and I'm trying my best to catch up with all of the contributor work.

I will be sure to update any documentation that refers to the current maintainers in the next few days.

andrewgodwin:
I'm also still around and helping out - it's not a dead project, as you suggest, just that life gets in the way sometimes. Work is underway on splitting out some parts to make things more generally useful, but it'll take a while - https://github.com/jointakahe/taktivitypub is the work in progress repo to split the AP parsing/generation code out, as that is one of the more complex pieces.

PR #713, opened 2024-05-21
Add myself to a few docs

AstraLuma:
Relates to #712


andrewgodwin:
If you want you can put both of us as authors  - I do want to come back and work on Takahē more now I've undone a lot of my burnout!

PR #722, opened 2024-08-28
Adjust dockerfile and docker compose to use environment variables throughout

davidfraser:
I'm not sure if this will meet your aims for docker and docker-compose.
But this allows me to put all the required usernames, passwords etc into a separate env file which I use from docker-compose to start a new development version of takahe. Otherwise there's duplication between docker-compose.yml and Dockerfile.


andrewgodwin:
The docker-compose file wasn't really meant to be anything other than just a way to get started on development - I'm reticent to add anything that may tempt people to use it for production...

davidfraser:
Fair enough...

PR #723, opened 2024-08-28
Minor python environment changes

davidfraser:
This:
* Substitutes [python-memcached](https://pypi.org/project/python-memcached/) (which is installable out of the box on Windows) for [pylibmc](https://pypi.org/project/pylibmc/) (which isn't), only on Windows
* Adds a git ignore for the common `venv` subdirectory for a working virtual environment when doing development


andrewgodwin:
Thanks!

